PR #6, opened 2014-01-19
Unify getPortName() across drivers

drewish:
On OS X if you open a virtual port using `input->openVirtualPort("foo bar")` then call `input->getPortName(1)` (assuming that's the right port number) it returns `foo bar`. 

Looking at the Windows code it seems to also just return the name.

Annoyingly, under ALSA it returns `RtMidi Input Client:0` making all but impossible to match to the name given when opening it. I modified this code to only return the port name.

I'm not quite clear what the Jack code returns.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -1423,23 +1423,13 @@ unsigned int MidiInAlsa :: getPortCount()
 
 std::string MidiInAlsa :: getPortName( unsigned int portNumber )
 {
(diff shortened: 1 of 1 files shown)


radarsat1:
This PR no longer applies cleanly and the conflict seems to affect the functionality of the PR.  I'd like to apply it but I'm not sure of the right fix.  Would anyone be willing to take a critical look at the port naming issues once again?

PR #27, opened 2014-05-03
Implement a reliable port selection API (official pull request against winks branch)

keinstein:
The current RtMidi implementation has a race condition that may be very annoying in certain cirumstances.

Steps to reproduce the problem:
1. add several virtual and/or usb devices
2. Start a RtMidi application with interactive port selection
3. After the Application presented the port list to the user, remove the removable device in the list from the system (delete or unplug it).
4. Select the a device that is below the unplugged device in the list.
5. RtMidi will fail or open the wrong device.

The current patch set contains a suggestion for a reliable API.
The main problem that shall be addressed is that the current API does not store port descriptors together with the meta data (port names/port numbers). Thus, both can get out of sync.

The patches are based both on master and winks include:

• Winks patches as well as reverting them to avoid problems synchronizing with the master branch.
• some code reorganization (extract common API into a separate class)
• Add a port descriptor class (With a more flexible port naming interface).
• Add port descriptor API: getPortList, openPort(PortDescriptor &, const std::string&), getPortDescriptor()
• Add Example programs.
• Add an automatic test case for all backends that support virtual devices. This has been inspired by the automatic loopback test case of the node-midi project.

I came across the following questions:
• What's the purpose of class RtMidi? Is it used directly?
• RtMidiIn and RtMidiOut serve as API containers. Why do you use virtual inline functions, here? I'd suggest to remove the virtual classifier from the member functions and declare them only in RtMidi if they are used both by RtMidiIn and RtMidiOut. This would render CommonMidiApi useless and I would reintegrate it into MidiApi. Otherwise the inline functions should be marked extern to avoid linker errors.
• There is no function call ot set or get the current API object from RtMidi(In|Out). This and the fact that the implementation classes Midi(In|Out)(Alsa|Jack|WinMM|WinKS|Core|Dummy) are not documented separately suggest that these classes are private to the RtMidi library. Is that right? In this case I'd suggest to remove these classes from RtMidi.h. Otherwise I'd suggest to add two functions to set and get the API object that is used by the RtMidi(In|Out).
• Maybe… probably… I might add a container API that allows to collect several APIs into one common API.


mbeards:
Is there still any interest in getting this PR merged?  The improved port selection API would be pretty helpful for me.

garyscavone:
This pull request involves a lot of changes and I haven’t had time to adequately evaluate it.  I was hoping the user community might provide some evaluation and feedback on the possible changes.  As always, the goal is to keep the code as simple, yet efficient as possible.

If people want to give the change a try and provide some feedback, that would help me out.

—gary

keinstein:
Sounds like an chicken and egg problem to me. I plan to use this API in my own projects and maintain the fork at https://github.com/keinstein/rtmidi until it can be merged. @mbeards , I think it would be helpful for all of us if you try this out and share your experiences with us, either as bug reports there or by commenting in this thread.

Regarding the code: 

The compatibility functions simply use the old code, so far. This is the efficient way, as it allows to open/close ports in O(1) on Windows or Mac OS (at least the user code part). The simple solution would provide system independent functions similar to those from the JACK backend, which have an open complexity of O(n) where n is the number of available MIDI devices. On windows this should not be a big issue as n cannot grow very large (at most 16, if I'm not mistaken). 

Another compitibilty issue is the usage of smart pointers. Even though the patch uses the ones that have been added to the C++ standard, not all compilers support this by default, yet. The compatibilty class can be removed if you drop the compatibilty with old compilers or (probably the better option) rely on Boost for the older C++ standards. The standard smart pointer API is a subset of the Boost API.

insolace:
Superseded by #30 and covered by #386

PR #52, opened 2015-08-05
Midiqueue mutex

morganpackard:
Hi there. This pull request isn't intended to ACTUALLY be pulled. It's just a convenient way to illustrate what I did to address some thread safety issues I was encountering on windows.


radarsat1:
Not for merge.

PR #55, opened 2015-12-04
Fix for closing virtual ports on OSX

szymonkaliski:
More info on this node-midi issue: https://github.com/justinlatimer/node-midi/issues/75

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -956,11 +956,17 @@ void MidiOutCore :: openPort( unsigned int portNumber, const std::string portNam
 
 void MidiOutCore :: closePort( void )
 {
-  if ( connected_ ) {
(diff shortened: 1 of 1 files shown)


garyscavone:
I don't have time to verify this is necessary ... the endPoint is not automatically disposed when the port is disposed? But if this is necessary, it seems likely it would also need to happen in the MidiInCore :: closePort() function?

szymonkaliski:
If you look at linked issue you'll see the problem — I'm opening four virtual ports and close them, but they all available in the system after that, this fixes it.

As for `MidiInCore :: closePort()` — I haven't looked there, but it might be needed as well.

szymonkaliski:
Midi In has the same problem, new PR: https://github.com/thestk/rtmidi/pull/57

PR #62, opened 2016-02-04
add travis script

MartinDelille:
Here is a *.travis.yml* script for *osx*. I tried to make it work for linux but the `autogen.sh` script failed I don't know why (maybe @radarsat1 has an idea?).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+os: osx
+language: cpp
+compiler:
+- clang
+- gcc
+before_script: ./autogen.sh
+script: make


SpotlightKid:
The problem with travis is that you can't load the Linux kernel midi modules. It's hard to do any functional testing without them.

radarsat1:
Can you elaborate on what didn't work with the autogen.sh script?  Perhaps on that pull request, or open an issue? We should solve it before Gary does a release.

MartinDelille:
sorry @radarsat1 I forgot the travis build link: https://travis-ci.org/MartinDelille/rtmidi/builds/107017758

MartinDelille:
@SpotlightKid I do plenty of functional test on OSX for my project: https://github.com/Phonations/Joker/blob/master/specs/MidiSpec/MidiSpec.cpp

SpotlightKid:
@MartinDelille Yeah, I was referring to Linux builds. Should have been more clear.

radarsat1:
Ah, it seems to have failed due to the weird %C% and %D% in the Makefile.am.  (This would have failed running automake by itself, regardless of autogen.sh...)

I don't know what those are for to be honest.  They seem to have appeared in commit 8b5fe770, reason stated as "rtmidi can now easily be nested into another project."  These variables seem to be a corner of automake that I'm not familiar with.  Perhaps an email to the author of that commit would be justified.

(For what it's worth, between Gary's testing and my own, they seem to cause errors when using automake-1.12 or older.  With automake-1.15 on my Ubuntu box, and when Gary used a recent automake from homebrew, these errors disappeared.)

MartinDelille:
@radarsat1 Since there is no way to contact the author (no github profile and no mail), maybe we can revert it?

radarsat1:
I've got a PR to revert just those changes, we are still discussing it.

radarsat1:
I'm not too familiar with Travis.  Is there a reason this Travis build is on Ubuntu 12.04, and gcc 4.6?  It seems that's why the automake is so out of date as to not support this feature of automake-1.14.  That said, it would be nice to add code to say "your automake is out of date, please upgrade!"

MartinDelille:
I made a try on ubuntu trusty 14.04 and have the following error:

    configure: error: ALSA sequencer support requires the asound library!

see full build log here: https://travis-ci.org/MartinDelille/rtmidi/builds/107332430

SpotlightKid:
Like I said, the travis Linux build environment can't load the ALSA kernel drivers, thatswhy the check for libasound probably fails.

MartinDelille:
@SpotlightKid ok sorry I thought the problem was about running test, not compilation.

SpotlightKid:
It's just a guess, to be sure you'd need to look at the `config.log` somehow.

MartinDelille:
The thing is that I used to compil rtmidi on my linux build: https://travis-ci.org/Phonations/Joker/jobs/107240449 (see install-7)

Here is my install script: https://github.com/Phonations/Joker/blob/master/scripts/install_rtmidi.sh

MartinDelille:
after some check, I can install *libasound2-dev* on precise where automake is out of date but no on trusty where it's unable to locate.

drewish:
I've been using Travis to test builds of a node.js module that wraps rtmidi: https://github.com/justinlatimer/node-midi/blob/master/.travis.yml

You should definitely be able to get it to compile on travis on Linux but you won't be able to run it since there's no dummy driver: https://github.com/travis-ci/travis-ci/issues/1754

garyscavone:
To be honest, I’ve never heard of Travis before (other than the 1-2 students that I’ve had with that name). :-) And I’m getting a bit frustrated with all this time and effort that is being spent on issues completely unrelated to the functionality of RtMidi. I just want a C++ class that provides useful MIDI functionality to users.

So, can we please try to wrap up issues related to building RtMidi early next week, so I can finalize a new release?  My standard priority is to keep things as simple as possible. Personally, I have no problem copying RtMidi.h/cpp to projects that need MIDI functionality, but I agree to let the library proponents get things in proper working order.

Regards,

—gary

radarsat1:
If PR#63 can be verified to configure and compile correctly I think we're good to go for a release. I don't think getting Travis working should be a release blocker.

MartinDelille:
@garyscavone [Travis CI](https://travis-ci.org/) is a continuous integration server that triggers build when a commit is pushed to *Github*. It allows *build matrix* in order to run several environment (linux/osx, clang/gcc, ...) and allow also to execute tests.

It'll make easy for you to check that a pull request doesn't break anything. Unfortunatly (as @SpotlightKid mentioned) there is no way to test midi driver on linux, but we could at least check the library compile fine.

My different try lead to the conclusion that we would need to make *rtmidi* compatible with automake 1.11.3.

What's your thought about it?

garyscavone:
I think the only things that are a problem for automake 1.11 are the “%D%”s in the Makefile.am, which are synonyms for "%reldir%”. Someone can first try replacing all instances of %D% with %reldir% to see if that works in 1.11. Otherwise, we can probably remove the %D%s altogether, as I’m not sure anyone is making use of that functionality.

radarsat1:
I agree that removing the %D% would be a perfectly good approach, that's why I proposed it before.. but on second thought I realized that if anyone is insisting on using an old automake while compiling from git instead of the release tarball, I think it's justified to just tell them to upgrade.  Older systems can compile from the .tar.gz (with generated configure) just fine.. that's the whole point of automake after all.

MartinDelille:
Could we advance with this issue? at least with the OSX environment? This would add reliability to the library and automate a part of the validation.

MartinDelille:
@garyscavone ?

garyscavone:
Vacation time … I probably won’t be able to look at this until late August.  And given that I know nothing about travis, it won’t be a fast process to review it.

MartinDelille:
There is not so much to review because the code impact is just a `.travis.yml` script configuring the service. Then you have to sign in under https://travis-ci.org and activate the RtMidi project in your account. I'll be happy to assist you doing these steps if you need help!

garyscavone:
To be honest, I don't expect I will use travis.  That said, it is not a big deal to have the .travis.yml script added to the repo. Are there other people out there interested in this?

SpotlightKid:
I'd vote for inclusion of the travis file.

drewish:
:+1: It's pretty handy when you get PRs from other people because you get a very early check if their changes will even compile.

MartinDelille:
@radarsat1 Any thoughts?

radarsat1:
Ok let's use it just for build verification for now, on RtMidi.  I'm using it on some other projects these days and it's useful.  However, I think both for RtMidi and RtAudio, it's too difficult to do any kind of functional testing on a VM, so let's not bother with using it for testing.

For the record, I think my and gary's objections are mostly the growing feeling that there is far too much infrastructure overhead growing around what amounts to a single cpp file with some examples.  Let's keep it simple, ok?  In any case, I think it's reasonable to use CI to at least check that "make;make install" doesn't get broken.

Another thing to please keep in mind, for anyone out there: if there _are_ breaking changes, look, this is the development repository.  If you need something stable please use the latest release.

PR #63, opened 2016-02-04
Install stuff

radarsat1:
This does a few things, maybe some debatable decisions, so I will list them clearly here:
  * Installs rtmidi.pc to /lib/pkgconfig so that pkg-config can find it.
  * Renames librtmidi.pc to just rtmidi.pc so that the cmd line is "pkg-config --libs rtmidi"
  * Installs header files on "make install"
  * Changes the header install path from /include to /include/rtmidi, updates rtmidi.pc and rtmidi-config to match.

This last change is just an idea.. it will of course break existing software if it doesn't use rtmidi-config or pkg-config in its build process.  I suggest testing a bit by compiling some user programs.


SpotlightKid:
All those things bring RtMidi more in line with the way things are done in other Open Source libraries so I personally would welcome them.

radarsat1:
Updated the PR.  It now includes building docs and tests from automake, and these are included in "make dist".

radarsat1:
Oh yeah, and docs are only built if doxygen is detected.  ("doxygen" executable must be available in PATH, or specified via DOXYGEN variable during configure)

radarsat1:
Just added a couple more:
  * prefer to run autoreconf if it's available, instead of running autotools manually
  * remove the %C% and %D% stuff from Makefile.am

For the latter, if we need it, it can be added back in future releases, but since it is causing problems, I propose remove it for this release.

garyscavone:
My guess is that the %C% and %D% stuff is to take input arguments? So, you could add a call to a larger project Makefile.am that calls the RtMidi configuration with particular paths?

radarsat1:
I think it must be something like that, but unless there's a way to disable it for older versions of automake I don't see what it adds.  I would say just remove it for now, we can always revert that commit for a later release.  I'll keep searching for some docs on that topic

radarsat1:
Oh I just noticed that adding the documentation to the build process has broken out-of-tree builds.  It should be an easy fix, just looking.

radarsat1:
Ok, I've fixed the documentation issues.  Left is the decision to leave %C% and %D%.  The author of that commit got back and it seems to be a feature of Makefile.am fragments introduced in automake-1.14, [documented here](https://www.gnu.org/software/automake/manual/html_node/Include.html).  It is not backwards compatible with older automake, so we just have to decide whether to support older automake, or support this feature of easily including RtMidi in other automake-based projects.  I would say that since more than one person has reported problems, it's probably not worth it for now, although the opposing argument is that those people should be using the latest stuff ;)

(Because, in principle, automake is only for maintainers... it's just that these days, people are installing things directly from github instead of downloading the tar releases!)

SpotlightKid:
How common is automake 1.14 in distro package repositories (taking the lifetime of LTS distros into account)?

radarsat1:
Seems that automake-1.14 is in Debian stable, that is good enough for me, personally.  Although it's confusing then why Travis doesn't build it properly.  I'll do a bit more research.

radarsat1:
I've updated the PR to remove the patch that removes %C% and %D%, and also to fix a problem with my documentation fix. ("cp -rl" doesn't not work on OS X)

radarsat1:
Switched from using cp to using AC_CONFIG_LINKS.  Seems to be more reliable.  Downside is that each file in doc/doxygen needs to be listed, but there aren't many.

radarsat1:
And now if I run with automake-1.11:

    $ automake-1.11
    configure.ac:6: require Automake 1.14, but have 1.11.6

I think this is a decent solution.  Opinions?

garyscavone:
This seems to work for me but I see that a make distclean leaves Makefile.in files in the doc/ and tests/ directories, as well as the main directory. Should those either be removed by the clean or added to the .gitignore file (which currently has /Makefile.in in it)?

radarsat1:
Ah yeah, I'll take a look..

radarsat1:
I think you proposed the right solution.  distclean normally leaves Makefile.in files, so they should just be ignored by git.

While checking this I noticed that "make distcheck" fails unless I "make" first, due to building the documentation.  Don't know if that's really an issue.. I'll just see if there's a quick solution.

radarsat1:
Ok, fixed make dist and make distcheck.

PR #68, opened 2016-02-16
Add error number when MidiInCore::initialize fails

MartinDelille:
In order to debug connection issue, I added the error code return by *MIDIClientCreate*. If needed I can add this behavior to other midi error case.


DouglasHeriot:
I’m interested – what was the issue you had, and how did you fix it?

I had an issue with MIDIClientCreate failing – I think it was caused by having multiple RtMidi instances with the same input port open. But I never got down to proving the cause. As soon as I consolidated some things and stopped creating multiple instances, the errors went away.

MartinDelille:
The issue is still there and still under investigation. I have the error code -50 return by *MIDIClientCreate* with correspond to *paramErr*. I suspect something with non ascii port name, is it your case to?

MartinDelille:
In fact the issue seemed to be related to multiple instance too!

MartinDelille:
What's wrong with this pull request?

PR #69, opened 2016-02-24
C api changes

remy-j-a-moueza:
I wanted to update my D bindings to the new RtMidi C api. 
However I had the following issues : 
- there was no handling of exception : my own C bindings captured the exception and the error message so that one could recover from an error or explain what was wrong; 
- in `rtmidi_in_get_message` the double return value of `RtMidiIn::getMessage` (number of seconds elapsed) was used to allocate the given `unsigned char ** message` array as if it was the length of the vector passed to `getMessage`; I needed a way to properly allocate the array and copy the data into a D GC allocated one before freeing the C malloc'ed memory.

So, I modified the C api as follow : 
- the RtMidiPtr `typedef`'s are now aliases to a simple structure : 

``` c++
struct RtMidiWrapper {
    void* ptr; // The RtMidi pointer
    bool  ok; // true if everything went right, false otherwise.
    const char* msg; // On error, the Exception .what() value.
};
```

the C api function capture the RtMidiError exceptions that may be thrown : they update the `ok` member to false and copy the exception `.what()` value to the `msg` member. 

The caller can inspect the structure, recover from the error and set back the `ok` member to `true` and `msg` back to the empty string before resuming or simply choose to exit the program. 
- `rtmidi_in_get_message` takes a third parameter : `size_t * size` that is used to tell the caller the size of the allocated array `unsigned char ** message`. When no message is received, no memory is allocated.

Doing so, I tried to minimize the changes to the API so that current users wouldn't have to change their code or would have a really minimum amount of changes to apply.  `rtmidi_in_get_message` is the only function which signature is modified.

### DIFF
--- a/rtmidi_c.cpp
+++ b/rtmidi_c.cpp
@@ -220,8 +220,6 @@ double rtmidi_in_get_message (RtMidiInPtr device,
         *size = v->size ();
 
         if (v->size () > 0) {
-            *message = (unsigned char *) 
-                        malloc (sizeof (unsigned char *) * (int) v->size ());
             memcpy (*message, v->data (), (int) v->size ());
         }
         delete v;
--- a/rtmidi_c.h
+++ b/rtmidi_c.h
@@ -13,17 +13,29 @@
 extern "C" {
 #endif
 
+//! Wraps an RtMidi object for C function return statuses.
 struct RtMidiWrapper {
+    //! The wrapped RtMidi object.
     void* ptr;
+
+    //! True when the last function call was OK. 
     bool  ok;
+
+    //! If an error occured (ok != true), set to an error message.
     const char* msg;
 };
 
+//! Typedef for a generic RtMidi pointer.
 typedef RtMidiWrapper* RtMidiPtr;
+
+//! Typedef for a generic RtMidiIn pointer.
 typedef RtMidiWrapper* RtMidiInPtr;
+
+//! Typedef for a generic RtMidiOut pointer.
 typedef RtMidiWrapper* RtMidiOutPtr;
 
-  enum RtMidiApi {
+
+enum RtMidiApi {
     RT_MIDI_API_UNSPECIFIED,    /*!< Search for a working compiled API. */
     RT_MIDI_API_MACOSX_CORE,    /*!< Macintosh OS-X Core Midi API. */
     RT_MIDI_API_LINUX_ALSA,     /*!< The Advanced Linux Sound Architecture API. */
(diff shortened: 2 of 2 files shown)


radarsat1:
I think the idea of capturing exceptions makes a lot of sense, I vote for this change.

emem67:
I have applied your change also to the callback proxy, I think it is more coherent, something like this:
midi_c.cpp
void callback_proxy(double timeStamp, std::vector<unsigned char> *message, void *userData)
{
	CallbackProxyUserData* data = reinterpret_cast<CallbackProxyUserData*> (userData);
	unsigned char* msg;
	if (message->size() > 0) {
		msg = (unsigned char *)malloc(sizeof(unsigned char *) * (int)message->size());
		memcpy(msg, message->data(), (int)message->size());
	}

	data->c_callback(timeStamp, msg, message->size(), data->user_data);
} 

midi_c.h
typedef void(*RtMidiCCallback) (double timeStamp, unsigned char* message, int size, void *userData);

what do you think?

remy-j-a-moueza:
It makes sense.
I overlooked `callback_proxy()` since I don't use callbacks to makes my C++/C/D bridge simpler.

I am not currently working on RtMidi related stuffs, so it may take a while before I test it and include it in a next commit. Or have you tested it already?

I have got a question related to memory management and lifetime.
Should the callback function be responsible deallocating the `malloc`'ed memory? It might be simpler for the user to have the memory freed after the call. The user would have to copy the `malloc`'ed C data s/he's interested in. Does this fit a usual data workflow? 
I am asking because I spent most of my times with garbage collected languages and may not have the right reflexes for C/C++.


```c++
void callback_proxy(double timeStamp, std::vector<unsigned char> *message, void *userData)
{
    CallbackProxyUserData* data = reinterpret_cast<CallbackProxyUserData*> (userData);
    unsigned char* msg;
    if (message->size() > 0) {
        msg = (unsigned char *)malloc(sizeof(unsigned char *) * (int)message->size());
        memcpy(msg, message->data(), (int)message->size());
    }
    data->c_callback(timeStamp, msg, message->size(), data->user_data);
   
    // Free the memory? The user should have made a copy of the data.
    free (msg);
}
```

emem67:
Hi,
I have tested the callback in my c# test app with and without the  
`  free (msg);`
and I have to admit you are 100% right :-) no memory leakage.
Thanks for the comment

radarsat1:
In principle it's a good idea to free memory that you allocate, unless it is very clear what the user should do.  However, in this case it's a high-frequency callback, it would be nice to avoid the malloc/free on every single call.. is it possible to keep the malloc'd data around?

On second look, is it necessary to allocate memory here?  Since message is an std::vector, it's perfectly valid to use its data pointer via data(), why not just pass that directly to the callback?

emem67:
@remy-j-a-moueza 
I have gone on with my experiments and in the c# application I have a memory leakage with the `rtmidi_in_get_message ` my guess is the c# application cannot free the memoty allocated vie the malloc.
Having the number of bytes returned variable I guess the correct approach would be to first get the number of bytes, then allocate a buffer and pass it to the `rtmidi_in_get_message `.
On the contrary the callback does not suffer of this problem.
@radarsat1 
I will make a test passing directly the vector data, in c# I use Marshal.Copy so it should be ok, I will report back.

emem67:
I have reverted back the callback to the original code by Atsushi Eno and it works well with my c# app without memory leakage.
So I vote to leave it as it is.
```
void callback_proxy(double timeStamp, std::vector<unsigned char> *message, void *userData)
{
	CallbackProxyUserData* data = reinterpret_cast<CallbackProxyUserData*> (userData);
	data->c_callback(timeStamp, message->data(), message->size(), data->user_data);
}
```
I think I have messed up a little with the c# wrapper so I got fooled.
I really apologize for wasting your time :-(

BTW I still have a memory leakage in `rtmidi_in_get_message` since I can not deallocate the memory in c# after it has been allocated by the malloc.
I will continue experimenting and in case I find something I will report here.
Many many thanks to both of you for the effort.

emem67:
I made a lot of test with the c# wrapper on the `rtmidi_in_get_message` and I have a (not so elegant) solution to prevent the memory leakage.
I simply allocate some unmanaged memory in c# then I call a modified `rtmidi_in_get_message` without the malloc then I free the memory in c#.
The `rtmidi_in_get_message` looks like this:
```
double rtmidi_in_get_message (RtMidiInPtr device, unsigned char **message, int *size)
{
	try {
		std::vector<unsigned char> *v = new std::vector<unsigned char> ();
		double ret = ((RtMidiIn*) device)->getMessage (v); 
		*size = v->size();

		if (v->size() > 0) {
			//*message = (unsigned char *)malloc(sizeof(unsigned char *) * (int)v->size());
			memcpy(*message, v->data(), (int)v->size());
		}
		delete v;
		return ret;
	} catch (...) {
		return -1;
	}
}
```

and the c# wrapper:

```
public byte[] GetMessage(out double timeStamp)
{
    IntPtr ptr = Marshal.AllocHGlobal(1024);

    int size;
    timeStamp = RtMidi.rtmidi_in_get_message(Handle, out ptr, out size);
    var buf = new byte[size];
    if (size > 0)
        Marshal.Copy(ptr, buf, 0, size);

    Marshal.FreeHGlobal(ptr);

    return buf;
}
```
The Marshal.AllocHGlobal(1024) does the memory allocation, I set 1024 bytes since it is the maximum defined for SYSEX messaged.
This is not so efficient but it works, I would recommend to use the callback in c#.
I have no idea if removing the malloc is good for the `rtmidi_in_get_message` I suspect it is a little bit too much c# oriented so maybe it should not be committed.

garyscavone:
There seem to be conflicts? Can someone fix this so it can be easily merged?

radarsat1:
@remy-j-a-moueza i fixed the conflict but i also added something that changes the semantics a bit of the "size" argument, please check.  (it now must contain the initial size of the "message" argument)  Otherwise it is a potential buffer overflow.

I recommend that you change this function also to return the "double" as an pointer-argument, since your -1 error code is mixed up with the "double" timestamp return value.

PR #95, opened 2016-10-25
changed printf to std::cout to fix compile error + make consistent wi…

johnty:
just checked out a clean copy of rtmidi and it fails to compile on test/midiclock.cpp in linux (both Raspbian and Ubuntu) due to to printf not being defined. it passes on OSX since its probably implicitly defined by the compiler.

since the rest of the program uses std::cout anyway, it probably makes sense to do so here as well?


radarsat1:
Thank you!  It's more a compiler version thing than Mac vs. Linux thing, since I developed it on Linux ;)

PR #108, opened 2017-04-21
MidiInAlsa::portInfo: don't filter SND_SEQ_PORT_TYPE_APPLICATION

jcelerier:
This patch makes PureData visible to RtMidi on Linux with alsa


jcelerier:
thanks!

PR #119, opened 2017-08-24
Compile fixes for WinMM

sagamusix:
* `INVALID_HANDLE_VALUE` is not compatible with `HMIDIIN` if `STRICT` is defined (which should always be the case, except for legacy code bases).
* Unconditional compile error in `midiInputCallback`.


sagamusix:
PS: In addition to Travis, you may want to look into AppVeyor for Windows CI, so that problems like this won't go unnoticed in the future.

Issue #118, opened 2017-08-21
API change to avoid unneccessary allocations


sagamusix:
This topic is intended to discuss possible API changes for reducing the amount of possibly required heap allocations, which are especially problematic in real-time contexts, which MIDI I/O typically happens in.
I would be tempted to start implementing these changes if this API and ABI change is welcome.

Rationale: Currently the RtMidi API enforces messages to be passed as a `std::vector<unsigned char>`. No matter what internal representation the API user may use in their own codebase, they have to convert it to a vector. Contrary to `std::string`, `std::vector` does not have any small vector optimizations, meaning that for every single message that is passed to `sendMessage`, a tiny bit of heap memory has to be allocated. Even if the user wants to use `std::vector` with a custom allocator, the custom allocator has to be dropped when passing the data to RtMidi, because the custom allocator causes the two vector variants to be different types. So an extra copy when converting a vector with a custom allocator to a standard vector cannot be avoided.

Solution: There are already lightweight alternatives to `std::vector` that could be adopted for RtMidi. For example, [gsl::span](https://github.com/Microsoft/GSL) is really easy to implement and may eventually become part of the C++ standard, in which case it could be easily dropped into the RtMidi code.
Essentially, a `span` is just two pointers that point to the beginning and end of some data, but it is not mandated where this data resides (inside some heap object, in an array on the stack, etc.). It's basically a view into any memory, such as the content of a `std::vector` (so existing code can easily be changed to use the new API).
I would be interested to add a new `sendMessage` overload to the C++ API to use something like `gsl::span` if it can be agreed that adding new API overloads makes sense. Replacing the existing `std::vector` functions would break existing code for no good reason, so it should simply forward to the new API.

It would also be good to know which C++ language standard should be aimed for when implementing this, as modern C++11/14 would certainly make things easier but as far as I can see, this is currently not a requirement for RtMidi. Bumping the required compiler version might thus not be a welcome change.

radarsat1:
I think a version taking std::array& would be welcome, but that also requires C++11.  I think probably a double* version is the best option.  I don't think adding a dependency on some 3rd-party library just for this is worth it however.

sagamusix:
[...]

PR #120, opened 2017-08-25
No allocations on sendMessage

radarsat1:
Add a second sendMessage prototype that takes a normal C array instead of std::vector.
From discussion in #118.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -2075,7 +2075,7 @@ static void CALLBACK midiInputCallback( HMIDIIN /*hmin*/,
   }
   else {
     // As long as we haven't reached our queue size limit, push the message.
-    if (!data->queue.push(message))
+    if (!data->queue.push(apiData->message))
       std::cerr << "\nMidiInWinMM: message queue limit reached!!\n\n";
   }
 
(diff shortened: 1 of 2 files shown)


garyscavone:
I'm just looking at this quickly. Should there be some documentation updates, since all the previous code examples and doxygen tutorial focused on the use of a vector?

radarsat1:
I don't think the examples need to change, the std::vector interface is still there. Of course it needs testing with stk, etc. Hopefully I'll find some time tomorrow.

garyscavone:
Thanks for the work on this! I'm going to go ahead and merge it. However, I don't quite understand the title. One way or the other, the user will have to allocate some memory to hold the MIDI bytes. And since the sendMessage() function took a reference, it wasn't performing any data allocation itself. So, I guess the concern here is that the std::vector constructor allocates too much memory? Aren't there ways to force the std::vector to minimize it's memory footprint (such as allocating it with a fixed size)?

sagamusix:
@garyscavone The initial motivation for writing this new overload is that my code doesn't use `std::vector` for MIDI message storage. So in the end, I had to allocate a temporary std::vector for every MIDI message to be sent. While a vector is perfectly suitable for storing MIDI messages, people may be using entirely different containers, and this pull request makes it possible.

sagamusix:
A perfect example for this is found in the C API wrapper for RtMidi by the way, which until now had to create a `std::vector` for passing the MIDI data on. Now that this pull request is merged, the C wrapper could be optimized to no longer allocate that vector.

radarsat1:
Thanks @garyscavone.  Just to answer your question, it's really a micro-optimisation but I think @sagamusix's point is not about the amount of memory or frequency of allocation, but (1) the ability to point to data on the stack instead of on the heap, and (2) the ability to directly provide sendMessage a pointer to memory in other container types without having to "convert" (i.e. copy) data to an std::vector.

PR #121, opened 2017-08-30
Ensure that port names are always UTF-8 on Windows, even when not compiling as UNICODE.

sagamusix:
As discussed at https://github.com/thestk/rtmidi/commit/6f710049e58507d3f4201c13af8962388a25fdfd


sagamusix:
I have no idea how it actually works but you may consider squashing those commits to get rid of some confusing commits. :)

radarsat1:
Okay I squashed and pushed it in 4adb2275fb7ae3bd6f4f591112a8aabfd5149cf1.
Made sure to mark you as the commit author.
Thanks!

PR #122, opened 2017-08-31
Static analyzer changes

rsjaffe:
Remove unnecessary null ptr check before delete.
change const std::string to reference for efficiency.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -12,7 +12,7 @@
 
     Permission is hereby granted, free of charge, to any person
     obtaining a copy of this software and associated documentation files
-    (the "Software"), to deal in the Software without restriction,
+    (the "Software" ), to deal in the Software without restriction,
     including without limitation the rights to use, copy, modify, merge,
     publish, distribute, sublicense, and/or sell copies of the Software,
     and to permit persons to whom the Software is furnished to do so,
@@ -61,8 +61,7 @@ RtMidi :: RtMidi()
 
 RtMidi :: ~RtMidi()
 {
-  if ( rtapi_ )
-    delete rtapi_;
+  delete rtapi_;
   rtapi_ = 0;
 }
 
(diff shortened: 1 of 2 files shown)


radarsat1:
Thanks, there are a few things going on here.

1. I think the most uncontroversial is a good catch for the unchecked result of openMidiApi, thanks for that, the `rtapi_` should indeed be null-checked.

2. The null checks before delete are harmless but I would accept it.

3. I'm not sure that passing std;:string by reference is still considered better, e.g. [this article](https://blogs.msdn.microsoft.com/xiangfan/2012/02/03/c98-c11-pass-by-value-or-pass-by-reference/)

4. There are a lot of weird and unnecessary whitespace changes, eg.,
~~~
-              CFStringAppend( result, CFSTR(", ") );
+              CFStringAppend( result, CFSTR( ", " ) );
~~~
or
~~~
-    ost << "MidiInCore::initialize: error creating OS-X MIDI client object (" << result << ").";
+    ost << "MidiInCore::initialize: error creating OS-X MIDI client object ( " << result << " ).";
~~~

what is the point of those kinds of changes?  There are even similar changes proposed in the comments and license header, which make no sense to me.

Can you make a separate PR or update this one to only fix the rtapi_ access?

rsjaffe:
Those whitespace changes were the result of my fighting against Visual Studio's attempts to reformat the code, and the ones you pointed out were in error. 

As to changing to const reference, see this in the CPP Core Guidelines document: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-in , which gives parameters for choosing between calling with a copy or with a reference. 

I can redo this pull request in a few days--traveling right now.

radarsat1:
Yeah, I guess the string gets copied two or three times in fact as it is passed from the constructor to openMidiApi to the subclass constructor, so in this case probably a reference is a good idea.  I *don't* think it will make any real difference in real performance of RtMidi, where typically the object is created once, or very rarely, but it's an acceptable "optimisation" I guess.

The whitespace changes are really pointless I'm afraid, I'm sorry your editor complains about it that's very silly.  If you don't mind, since you are travelling I'll cherry-pick the other changes myself put you as commit author.  Only because we are doing a new version *today* so we may as well include the function signature change in the version bump.

radarsat1:
Push with some fixes in b0cacfdc2f31f831e6cfabd0e1c282a4befaf81b

PR #123, opened 2017-09-01
C API: Make use of new sendMessage overload

sagamusix:
Directly pass on the pointer+size for message instead of creating a superfluous vector.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -47,8 +47,8 @@
 #endif
 
 // Default for Windows is to add an identifier to the port names; this
-// flag can be undefined to disable this behaviour.
-#define RTMIDI_ENSURE_UNIQUE_PORTNAMES
+// flag can be defined (e.g. in your project file) to disable this behaviour.
+//#define RTMIDI_DO_NOT_ENSURE_UNIQUE_PORTNAMES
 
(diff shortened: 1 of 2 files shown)


sagamusix:
I hope you don't mind that I added another commit to this pull request; The second commit inverts the behaviour of commit 26c77d45e0502b78f51aa8d244c5cb791978e5bd.
Rationale: It is possible to add a macro definition "from the outside" without having to touch the library code, but it is possible to undefine a macro that is later begin defined in the code. So with the inverted mechanism, it is only necessary to edit the RtMidi project file (the official project files could later be modified to have a separate configuration for this) rather than having to edit the .cpp file with every new release.

radarsat1:
Hi, I am applying this, but I just want to say that I wish you had made that second patch a separate PR.  However, I think that mechanism needs an overhaul anyway and is subject to change so I won't make a fuss.  (We need to figure out a general sane "port names" strategy that works for all APIs, so likely this flag will disappear.  But let's keep that discussion in an issue.)

sagamusix:
I keep forgetting that a pull request on GitHub includes all commits of a branch so by the time I committed the second commit, it was too late. I guess I'm too used to .patch files. ;)

PR #126, opened 2017-09-29
 Build breaks when compiling with --prefix and upgrading version.

amosonn:
Removed line which adds, when configured with prefix, include to the currently installed version (assuming prefix is always the same).
This include causes submodules (i.e. tests) to link against the currently installed library, instead of the newly compiled one, breaking build when upgrading version.


radarsat1:
Thanks. This makes some sense.  I was wondering if `-I$(top_srcdir)` is needed instead? Or is the behaviour of AM_CPPFLAGS sufficient?  https://www.gnu.org/software/automake/manual/html_node/Program-Variables.html

Not sure why you are closing and opening this.

radarsat1:
Okay I see, just leave it. I'll post my question in #127.

PR #127, opened 2017-09-29
 Build breaks when compiling with --prefix and upgrading version.

amosonn:
Removed line which adds, when configured with prefix, include to the currently installed version (assuming prefix is always the same).
This include causes submodules (i.e. tests) to link against the currently installed library, instead of the newly compiled one, breaking build when upgrading version.

(I managed to break last pull request, opening new one. Sorry.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -77,7 +77,6 @@ AC_ARG_ENABLE(debug,
 # Set paths if prefix is defined
 if test "x$prefix" != "x" && test "x$prefix" != "xNONE"; then
   LIBS="$LIBS -L$prefix/lib"
-  CPPFLAGS="$CPPFLAGS -I$prefix/include/rtmidi"
 fi
 
 # For -I and -D flags


radarsat1:
Thanks. This makes some sense. I was wondering if -I$(top_srcdir) is needed instead? Or is the behaviour of AM_CPPFLAGS sufficient? https://www.gnu.org/software/automake/manual/html_node/Program-Variables.html

amosonn:
I think that `-I$(top_srcdir)` exists in `AM_CXXFLAGS`, and that is enough; it's just that `CPPFLAGS` appears earlier in `CXXCOMPILE` and `LTCXXCOMPILE`, and therefore the search was first done in `$prefix/include/rtmidi`.
Anyway, it compiles after this fix.

radarsat1:
Thanks! I think that must be left over from before we used automake.

amosonn:
Actually, it seems that `AM_CXXFLAGS` with `-I$(top_srcdir)` appears only in `tests/Makefile.am`, not in the root `Makefile.am`. I don't know if this is a problem.

radarsat1:
No I believe that should be correct, only the tests need access to RtMidi.h in the srcdir.  Maybe I'm wrong about the reason that's there ;)  but good catch, I think your change is correct.

amosonn:
Glad to help :)

PR #129, opened 2017-10-03
C API improvements/fixes

zserge:
This covers the issues with the C API that I've found when working on Go bindings. Most of them look critical to me, so I dared to make some changes:

1. Replace `typedef RtMidiWrapper* RtMidiPtr;` with `typedef struct RtMidiWrapper* RtMidiPtr;` because the former is invalid C code and doesn't compile in C99 mode.

2. Removed RT_MIDI_API_WINDOWS_KS from the enum to make C API enum absolutely match RtAudio C++ API enum. Otherwise the numeric values will be incorrect. I haven't found any uses of RT_MIDI_API_WINDOWS_KS in the code, so I assumed it's safe to remove it from the C API, if it's not in the C++ API.

3. Removed rtmidi_sizeof_rtmidi_api. Well, this is a matter of taste, but I find `sizeof(enum RtMidiApi)` to be much more readable and explicit. Since the type is defined in the header I don't see any reasons to make a helper function for its size.

4. Removed double pointers in `rtmidi_in_get_message` and `rtmidi_get_compiled_api`. These functions do not allocate pointers, they only have been taking double pointer to double-unreference it. I find it much more clear to pass a regular pointer instead.

5. In the midi callback there was no message size. And it seems to be rather impossible to guess how may bytes from `void *message` it's safe to read without crashing. So I've changed the callback prototype to have a message size in bytes.

That being said, with these changes I find C API to be usable, and I'd be happy to add Go bindings for this library in a manner I'm doing now with RtAudio (https://github.com/thestk/rtaudio/pull/99).

Ah, and thanks for the wonderful libraries!

### DIFF
--- a/rtmidi_c.cpp
+++ b/rtmidi_c.cpp
@@ -14,12 +14,6 @@ class CallbackProxyUserData
 	void *user_data;
 };
 
-/* misc */
-int rtmidi_sizeof_rtmidi_api ()
-{
-	return sizeof (RtMidiApi);
-}
-
 /* RtMidi API */
 int rtmidi_get_compiled_api (enum RtMidiApi **apis) // return length for NULL argument.
 {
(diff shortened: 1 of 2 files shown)


radarsat1:
All sounds reasonable to me, thanks.

PR #138, opened 2017-10-19
Removed extern

micdah:
RtMidi does not compile on Windows when `extern` modifiers on constructors are added

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -124,7 +124,7 @@ void RtMidiIn :: openMidiApi( RtMidi::Api api, const std::string &clientName, un
 #endif
 }
 
-extern RTMIDI_DLL_PUBLIC RtMidiIn :: RtMidiIn( RtMidi::Api api, const std::string &clientName, unsigned int queueSizeLimit )
+RTMIDI_DLL_PUBLIC RtMidiIn :: RtMidiIn( RtMidi::Api api, const std::string &clientName, unsigned int queueSizeLimit )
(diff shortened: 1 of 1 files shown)


radarsat1:
Ah, it would be nice if this works better than

```
git revert 0470e7c5e5fc70ca22351883cab76c8e1222438a
```

but I'll have to test.  Thanks!

micdah:
Well actually these are two different problems I have encountered. This one is about being unable to compile on Windows.

The other with regards to reverting the above commit, is about not being able to use the compiled `.dylib` on macOS when using the methods via P/Invoke from .Net Core.

radarsat1:
aaaah ok, thank you, I thought they were related.  (Or at least, I am working too fast on multiple things atm and getting confused.. XD)

Again, I'll test this and consider merging.  Which compiler have you tested?

micdah:
Sorry for the confusion, I feel you, I am extremely bad at multi-tasking myself 😃

umlaeute:
btw, i think that the `extern` keywords before the constructors where accidental and shouldn't have been there in the first place. so from the Debian side: please merge this

micdah:
Forgot about the compilers, I am using the default compiler provided with macOS (I have Xcode developer tools installed), and the last dependencies (automake, autoconf) I have installed using Homebrew.

radarsat1:
Rebased and pushed in bb5f772c4baa1a2744ef4ac6a64b35fc6842b97c

Issue #137, opened 2017-10-19
Cannot find entry point (C API macOS)


micdah:
I am having difficult compiling rtmidi on macOS after recent changes to `Makefile.am` and `configure.ac`, it seems the exposed C methods are being garbled where they weren't previously (_I am using rtmidi from managed code via P/Invoke_).

If I revert these to files to there state in commit `d19b7573f7274edbfe8ece52fae482c66094d767`, I have no issue compiling and using the C-methods from managed code.

This is how I compile rtmidi:
```
# ./autogen.sh --no-configure
# ./configure
# make
```

I am using the produced `.libs/librtmidi.dylib` file.

radarsat1:
Can you possibly do a bisect and find where the problem occurs?  The commit you indicate only changes some lines in `README.md`.

micdah:
Oops forgot to mention, the error I am receiving is _"Unable to find entry point `method_name` in DLL rtmidi"_

micdah:
I suspect it is one of the commits after that point (there are a handfull recently changing both these files), so my debugging only led me to try and specifically pick these version `git checkout d19b7573f7274edbfe8ece52fae482c66094d767 Makefile.am` (and likewise for `configure.ac`) and after doing this, I can observe that I no longer have this issue and rtmidi seems to function correctly in my application.

I can dump a diff here, if that helps?

There are a great deal of changes, so I am having a difficult time pinpointing the smoking gun, so to speak.

**Makefile.am diff**
```
diff --git a/Makefile.am b/Makefile.am
index 2bbb17c..1606cc7 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,8 +3,6 @@ if MAKE_DOC
SUBDIRS += doc
endif

AM_CXXFLAGS = @visibility@

lib_LTLIBRARIES = %D%/librtmidi.la
%C%_librtmidi_la_LDFLAGS = -no-undefined -export-dynamic -version-info @SO_VERSION@
%C%_librtmidi_la_SOURCES = \
```

**configure.ac diff**
```
diff --git a/configure.ac b/configure.ac
index 4390648..2474c4c 100644
--- a/configure.ac
+++ b/configure.ac
@@ -27,49 +27,31 @@ m4_define([lt_current_minus_age], [m4_eval(lt_current - lt_age)])

SO_VERSION=lt_version_info
AC_SUBST(SO_VERSION)
AC_SUBST(LIBS)
AC_SUBST(api)
[...]

PR #139, opened 2017-10-19
Revert "Enabled visibility=hidden for non-debug builds (only)"

micdah:
This reverts commit 0470e7c5e5fc70ca22351883cab76c8e1222438a.

I have found that with this change, the resulting `.libs/librtmidi.dylib` doesn't seem to expose the C-API methods when being invoked using P/Invoke from .Net Core managed code.

Reverting this specific commit, seems to "fix" the issue. As after this, my code again starts working and P/Invoke again is able to invoke the methods.

See this issue for further discussion https://github.com/thestk/rtmidi/issues/137

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,8 +3,6 @@ if MAKE_DOC
 SUBDIRS += doc
 endif
 
-AM_CXXFLAGS = @visibility@
-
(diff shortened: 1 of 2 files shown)


radarsat1:
Rebased & pushed in 9eb59e4839f8e230a496a45e3e7e1b5e7570c7fe

PR #141, opened 2017-11-05
Use override designation when appropriate

rsjaffe:
Remove redundant virtuals

Reduces risk of error when overriding virtual functions


radarsat1:
Hi, I'm closing for now, but I'll re-open this when we switch to C++11.

PR #142, opened 2017-11-05
Used nullptr instead of 0 or null where appropriate

rsjaffe:
Did not convert over OSX code, as Apple documentation is unclear as to whether nullptr would be accepted.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -55,14 +55,14 @@
 //*********************************************************************//
 
 RtMidi :: RtMidi()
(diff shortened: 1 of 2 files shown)


rsjaffe:
Looks like the Travis versions of the compilers are a bit old and don't support nullptr. Close this pull request if you don't want the hassle of dealing with that issue.

radarsat1:
Hi, I'm closing for now, but I'll re-open this when we switch to C++11.

Issue #136, opened 2017-10-10
hide private classes


umlaeute:
it seems that the backend classes are meant to be a private implementation details, and should not be exposed in `RtMidi.h`

umlaeute:
i tried to search github repositories, and the only uses of `MidiInJack` i found so far were verbatim copies of `RtMidi.cpp`, `RtMidi.h` or its documentation.

However, after wading through 400 repositories (with 1718 hits)  I managed to hit the rate-limit of github (and trigger an abuse detection mechanism), so the search was not exhaustive yet.

I also did a (possible not exhaustive) search on https://codesearch.debian.net with similar results.
I'm pretty sure that these classes are not used directly by any consumer.

umlaeute:
ah, and my recent `visibility` fixes (merged into both RtAudio and RtMidi) might actually prevent the use of the backend implementation classes for non-debug builds anyhow (the backend classes are *not* marked as being exported)

radarsat1:
Fixed in #158, I think.

PR #158, opened 2018-03-28
Move MidiInApi and MidiOutApi subclasses into RtMidi.cpp

radarsat1:
Addresses #136.
@garyscavone any issues with this?  It seems @umlaeute has done the adequate research on reverse dependencies.
The idea is just to not expose implementation classes in the header, to ensure that code using the library cannot depend on specific backend implementation classes.

I could not succeed in hiding RtMidiApi due to the inline definitions which require the class to be fully defined.


garyscavone:
That makes sense to me … not sure why I didn’t do it that way when I first created RtMidi.

FYI, my original plan when creating RtMidi was that it could function as input, output or both in/out in a single class (like RtAudio). But as I started to develop it, the input and output functionalities were so distinct that I couldn’t come up with an API that I liked.

radarsat1:
Ah, yeah I understand that.  There is no "duplex" API, I guess, but I guess that's ok for now.  It still makes send to have a base class for handling devices etc.  Ok, I'll merge this, thanks for the review.

PR #159, opened 2018-04-01
add setPortName() functionality

chronopoulos:
This commit adds a new method setPortName(const std::string &portName)
to the RtMidi class. Port renaming is implemented for the ALSA and
JACK API's. For the other API's, a warning is thrown using RtMidiError.

Note that port renaming in JACK is implemented using jack_port_rename(),
which is available on jack2 versions >= 1.9.11, and jack1 versions >=
0.125.0.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -3099,7 +3099,11 @@ void MidiInJack:: setClientName( const std::string& )
 void MidiInJack :: setPortName( const std::string &portName )
 {
   JackMidiData *data = static_cast<JackMidiData *> (apiData_);
+#ifdef __JACK_HAS_PORT_RENAME__
   jack_port_rename( data->client, data->port, portName.c_str() );
+#else
+  jack_port_set_name( data->port, portName.c_str() );
+#endif
 }
 
 //*********************************************************************//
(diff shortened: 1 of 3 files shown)


chronopoulos:
well, i somewhat expected this would happen. the issue is that for old versions of jack (1 and 2), port-renaming is done using jack_port_set_name(). this function is now deprecated, so this commit uses the newer jack_port_rename() instead. i guess some build tests are using the older libraries.

how can we detect library versions (or the existence of a function) using pre-processor directives so that RtMidi can continue to work with a variety of build systems?

radarsat1:
Ah, ok.  This kind of thing is typically handled via autoconf [AC_CHECK_LIB](https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Libraries.html), which allows you to check that a library has a function.  It looks like the equivalent in cmake is [CheckLibraryExists](https://cmake.org/cmake/help/latest/module/CheckLibraryExists.html)

chronopoulos:
ok thanks, i added a line in configure.ac which checks for the existence of jack_port_rename. build tests are passing now.

chronopoulos:
@radarsat1 anything holding up this merge that i can help with? its scope is very similar to the setClientName branch that was merged two weeks ago (thanks for that!)

PR #160, opened 2018-04-02
[WIP] Add CMake support

jgvictores:
All examples tested and working with physical and virtual devices on Ubuntu Xenial 16.04 (gcc 4.5, cmake 3.5.1, ALSA) and Windows 10 (VS 15 2017, win32, cmake 3.9.4, WinMM). Prepared for Mac, but untested.

Once compiled (generates the C++ static library) and installed (header also installed), the required `CMakeLists.txt` for consumer programs is minimal:
```cmake
find_package(RtMidi REQUIRED)
add_executable(example example.cpp)
target_link_libraries(example RtMidi::rtmidi)
```
And the generated solutions/makefiles need no additional tuning, with all paths set automatically on the tested platforms.


radarsat1:
Hi, thank you, looks nice.  Some things missing:

* default to shared lib (`set(BUILD_SHARED_LIBS ON)`)
* set version=4, soversion=4.0.0 
* build tests and tools
* generate & install pkg-config file
* ability to select JACK

jgvictores:
I'll be writing some notes here, if you don't mind, just to track progress and not open a new issue.
- Adding version is easy and nicely done via [PROJECT VERSION](https://cmake.org/cmake/help/v3.0/command/project.html) and setting target properties ([example](https://github.com/roboticslab-uc3m/amor-api/blob/7ed092c7f6daf84ebab8dc41f855e940aee281d8/libraries/amor/CMakeLists.txt#L47-L50)).
- Other points seem feasible.
- Having some issues with shared lib in Windows. When passing to SHARED, the `.dll` is compiled and installed. However, consumers cannot compile due to a missing `build\Release\rtmidi.lib`. Seems strange, but according to [this old thread](https://cmake.org/pipermail/cmake/2009-November/033331.html) it could be comprehensible: `You don't link against a dll. You link against the import library (.lib) and then the DLL must be available to your application at runtime.`

radarsat1:
It should be possible to make use of RTMIDI_DLL_PUBLIC macro in RtMidi.h for declaring exported symbols via `__declspec( dllexport )`

PROJECT_VERSION is for the project version (currently 3.0.0), but what I am proposing is to add the VERSION and SOVERSION to the target properties, which corresponds with the ABI version, currently 4.0.0.  They are not the same, e.g. RtMidi could release 3.1 which removes a function, bumping the ABI version to 5.0.0.  This is mainly for distributions that are concerned with binary compatibility, but it would be nice if this were synchronized correctly between autotools and CMake.  Bonus points if CMake can read the info out of `configure.ac`.

ps. yes definitely feel free to track your progress here ;)

radarsat1:
For the record: I'd actually prefer not to have two build systems but so many people prefer CMake now that we may as well accept this work, as we have for [RtAudio](https://github.com/thestk/rtaudio/blob/master/CMakeLists.txt); in that case I'm just insisting that they are at least equivalent.

PeterBowman:
> It should be possible to make use of RTMIDI_DLL_PUBLIC macro in RtMidi.h for declaring exported symbols via __declspec( dllexport )

CMake could take care of this by means of the [GenerateExportHeader module](https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html). In fact, the following snippet could be replaced by a header inclusion (in the CMake world):

https://github.com/thestk/rtmidi/blob/22cc8b3aff966e61c0aed1361d3fb96462cd9b50/RtMidi.h#L46-L54

> Bonus points if CMake can read the info out of configure.ac.

I'm afraid the migration effort is a manual process. Here are some guidelines: https://stackoverflow.com/a/7680240.

radarsat1:
> CMake could take care of this by means of the GenerateExportHeader module.

We are not dropping autotools for CMake, so that's not a possible solution.  It's trivial to add a couple of lines to that snippet to add `__declspec(dllexport)` in the Windows case.

> > Bonus points if CMake can read the info out of configure.ac.
> I'm afraid the migration effort is a manual process. Here are some guidelines: https://stackoverflow.com/a/7680240.

I didn't mean to do anything automatically. I meant writing CMake code to read the SOVERSION from configure.ac since the information is already there. It could be extracted via grep for instance, but probably a Windows-portable solution would be preferable.

jgvictores:
> Perhaps WINDOWS_EXPORT_ALL_SYMBOLS, introduced at CMake 3.4, could help

It turns out that CMake `WINDOWS_EXPORT_ALL_SYMBOLS` as implemented [here](https://github.com/jgvictores/rtmidi/compare/cmake-support...jgvictores:332200a) works marvellously. I really prefer this solution because it allows the cmake to be more self-contained, in the sense of not having to touch the header ever at all.

Note: Enabling SHARED on the Windows 10 setup described in the issue description leads to a bad mix of 32 and 64-bit linked libraries (thank you [depends.exe](http://www.dependencywalker.com/)) Surprisingly (at least to me, given the project description), everything works correctly (tested WinMM input/output) by setting the compiler to 64-bit mode.

PR #163, opened 2018-04-24
Add optional exception support. (__NO_EXCEPTIONS__ for -fno-exceptions and -fno-rtti)

jwurzer:
Hi, I have to compile RtMidi without exceptions because the projects where it should be used uses -fno-exceptions and -fno-rtti.
Therefore I changed RtMidi with an optional Option to disable exceptions.

If you want to compile RtMidi without exceptions support then the macro `__NO_EXCEPTIONS__` must be defined and the correct compiler options (e.g. -fno-rtti and -fno-exceptions) must be used. Every time when the header file RtMidi.h is used then the macro `__NO_EXCEPTIONS__` must be defined. Otherwise the definitions inside RtMidi.h are not compatible with the compiled version.

Instead by an exception an error can be recognized with RtMidi::isError(). See rtmidi_c.cpp as example to use RtMidi without exceptions. Only RtMidi.cpp and rtmidi_c.cpp are changed to compile without exceptions but not the tests-apps under tests/.

Maybe you find it useful to include the optional exception support into RtMidi.


samhocevar:
I would like to provide two comments about this patch:

 - if the `THROW` etc. macros are to be defined in `RtMidi.h` then they should probably be called `RTMIDI_THROW` etc. so as not to mess with macros that the user’s project defines.
 - you can probably avoid the code duplication caused by `try`/`catch` by using these macros: `#define TRY /* */` and `#define CATCH(x...) for(x...;false;)`

garyscavone:
I'm not opposed to removing exceptions in RtMidi but this PR is so old that I don't expect it is useful any longer.

PR #173, opened 2018-09-03
cmake: Propagate API defines for test targets

pthorod:
Needed to compile out of the box on windows, ex '#if defined(__WINDOWS_MM__)' in midiclock.cpp


radarsat1:
Since these defines aren't part of the API it's better to fix the tests to not need them.

PR #174, opened 2018-09-05
Fix some bad timestamps on ALSA

jpcima:
The ALSA interface emits some very high timestamps which are produced as a result of integer overflow in the calculation.
By playback of large numbers of MIDI files into RtMidi's ALSA client, I find such occurrences of the issue to happen extremely frequently, usually within first seconds of playback.

This is the reason: a computation applies a substraction principle from `timespec` values originating from GNU documention. Probably nothing is wrong with this method, but this is the deal. Compare:

```
typedef struct snd_seq_real_time {
        unsigned int tv_sec;            /**< seconds */
        unsigned int tv_nsec;           /**< nanoseconds */
} snd_seq_real_time_t;
```

```
struct timespec {
  time_t tv_sec; /* Seconds.  */
  long tv_nsec;  /* Nanoseconds.  */
};
```

As you can observe, ALSA's variant of timespec stores its nanosecond time as unsigned value.
When the substraction method puts a negative in the nanosecond field, it usually overflows into unsigned value in the measure of billions, which doesn't do well in the timestamp computation following.

I think a simple fix can be as proposed, doing all the computation in floating points where is eliminated any possibility of overflow.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -1600,27 +1600,13 @@ static void *alsaMidiHandler( void *ptr )
           // Method 2: Use the ALSA sequencer event time data.
           // (thanks to Pedro Lopez-Cabanillas!).
 
-          // Using method from:
-          // https://www.gnu.org/software/libc/manual/html_node/Elapsed-Time.html
-
-          // Perform the carry for the later subtraction by updating y.
-          snd_seq_real_time_t &x(ev->time.time);
-          snd_seq_real_time_t &y(apiData->lastTime);
-          if (x.tv_nsec < y.tv_nsec) {
-              int nsec = (y.tv_nsec - x.tv_nsec) / 1000000000 + 1;
-              y.tv_nsec -= 1000000000 * nsec;
-              y.tv_sec += nsec;
-          }
-          if (x.tv_nsec - y.tv_nsec > 1000000000) {
-              int nsec = (x.tv_nsec - y.tv_nsec) / 1000000000;
-              y.tv_nsec += 1000000000 * nsec;
-              y.tv_sec -= nsec;
-          }
+          snd_seq_real_time_t x = ev->time.time;
+          snd_seq_real_time_t y = apiData->lastTime;
 
           // Compute the time difference.
-          time = x.tv_sec - y.tv_sec + (x.tv_nsec - y.tv_nsec)*1e-9;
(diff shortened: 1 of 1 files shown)


radarsat1:
Thanks for reporting this!
I am not sure floating point calculation are the correct fix, since it could lead to imprecision. Isn't there a way to fix it with larger integer calculations?  It would be good to see a way to test this independent of feeding in random MIDI files.  Some unit testing would be beneficial here.

jpcima:
>I am not sure floating point calculation are the correct fix

The double precision epsilon raises over 10^-9 starting at the decade of 2^23, 10^-6 at 2^33, and 10^-3 at 2^43. Meaning, the float resolution can be approximated to become less precise than the nanosecond after 2330 hours of accumulated timestamp.
That's not accounting for the accumulated computation error but still, I think of doubles as largely reasonable.

>It would be good to see a way to test this independent of feeding in random MIDI files

Let's see if I can isolate a sequence which triggers the problem

:warning: This said, I believe there may be a greater problem about the timestamps. As some events are bypassed because of ignored types, their timestamps are going to be also discarded. I would think that these ignored timestamps must be kept accumulated for the next event.

radarsat1:
> Let's see if I can isolate a sequence which triggers the problem

Oh don't worry about that, I think it's best to just test the timestamp code in isolation.  I'll do it as soon as I get a few minutes.  Again, thanks for reporting this!

jpcima:
@radarsat1 If the target has `sizeof unsigned int`=`sizeof long`=4, the substraction of `nsec` fields implicitly converts operands to unsigned types.
I believe on Linux 32-bit OS, the patch 6d01662c9ea46a43ef1719d3464d8fc0db1469c6 will still yield incorrect result.

radarsat1:
You were right! My test program failed when compiled for 32-bit. Actually I found only the only calculation that made a problem was x.tv_sec field leading the calculation for `time`, and just re-ordering this fixed it, but I added `(int)` casts instead to be more explicit.  Thanks!

radarsat1:
btw I was trying to imagine whether it needs to be `(long long)` instead of `(int)`, due to losing a bit in the signed conversion, but it didn't seem to trigger any issues using just `(int)`.  My test program iterates over all values of `tv_sec` and `tv_usec` for both variables and checks for differences in signed or unsigned calculations.

PR #180, opened 2018-10-21
Fix the input of system-exclusive messages on ALSA

jpcima:
This repairs a problem introduced by commit 3155671
"alsa: fix a switch fallthrough warning", which makes it impossible to
receive any system-exclusive message.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -1567,6 +1567,7 @@ static void *alsaMidiHandler( void *ptr )
           break;
         }
       }
+      doDecode = true;
       break;
 
     default:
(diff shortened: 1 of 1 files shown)


SpotlightKid:
Maybe it would be a good idea to fix the wrong indentation of the case statement above as well, which maybe contributed to this error.

jpcima:
Yes I think, so I've added this change too.

jpcima:
Ping. Can I do something in order to help include this fix?
The software [FreeMajor](https://github.com/linuxmao-org/FreeMajor/) has dysfunctional ALSA input on vanilla RtMidi.

PR #186, opened 2019-01-07
 Allow WinMM users to select RT_SYSEX_BUFFER_SIZE for each port.

asarhaddon:
Hello.
The two first commits are only normalizing spaces and should be consensual.
The last one does what the title suggests.
SpotlightKid also suggests that the new function becomes a method only affecting the WinMM API and doing nothing on other APIs,
I do not mind the procedure as long as it is possible to set the buffer size at run time.
Thanks.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -2386,7 +2386,10 @@ static std::string ConvertToUTF8(const TCHAR *str)
   return u8str;
 }
 
-#define  RT_SYSEX_BUFFER_SIZE 1024
+static size_t RT_SYSEX_BUFFER_SIZE = 1024;
+void setRTSysExBufferSize( size_t size ) {
+  RT_SYSEX_BUFFER_SIZE = size;
+}
 #define  RT_SYSEX_BUFFER_COUNT 4
 
 // A structure to hold variables related to the CoreMIDI API
--- a/RtMidi.h
+++ b/RtMidi.h
@@ -598,6 +598,16 @@ class RTMIDI_DLL_PUBLIC MidiOutApi : public MidiApi
 
 // **************************************************************** //
 //
+// On Windows with the WinMM API, set the size of input buffers
+// created by subsequent calls to openPort(). The default is 1024.
+//
+// **************************************************************** //
+#if defined(__WINDOWS_MM__)
+void setRTSysExBufferSize( size_t size );
+#endif
+
+// **************************************************************** //
+//
 // Inline RtMidiIn and RtMidiOut definitions.
 //
 // **************************************************************** //
--- a/doc/doxygen/tutorial.txt
+++ b/doc/doxygen/tutorial.txt
@@ -397,7 +397,7 @@ The RtMidi JACK support can be compiled on Macintosh OS-X systems, as well as in
 
 The \c configure script provides support for the MinGW compiler.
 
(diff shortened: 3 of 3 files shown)


garyscavone:
I would merge the first two commits right away but I can't see how to cherry-pick from this web interface and I don't want to mess things up. The third commit involves an API change and should be submitted separately so that the changes can be more easily analyzed.

SpotlightKid:
The first two commits are probably obsoleted by the already applied PR #187 anyway, so @asarhaddon should rebase his branch.

asarhaddon:
I have rebased the 3 short patches.
After some time trying to upload the changes to github, I am now giving up and sending the patches as attachments to the email address found in configure.ac.
Feel free to contact me if I can help.

SpotlightKid:
I you rebase, you must push your branch with ``git push --force-with-lease=<your branch name>``.

garyscavone:
I’m not comfortable dealing with individual patch files. Please rebase and then resubmit the PR. Also, keep the whitespace / tab issues in a separate PR from the sysex ones.

radarsat1:
I am able to cherry-pick the sysex patch on top of master just fine, however I think the proposed change is not very good since it adds a global function.  It is fine for your personal use of course, but as an addition to the library it is quite hacky.  @garyscavone  what do you think about settable buffer size for sysex?   If it's a general need it could be added to RtMidiIn more cleanly.  (It's only for input, is that correct?)

Does anyone know _why_ this is specific to WINMM, do other APIs not have to deal with buffer sizes for sysex?

asarhaddon:
@radarsat1
The commit header explains why no default can possibly satisfy the whole range of hardware needs.

Hopefully you agree a manual edition of the header each time RTMIDI is updated is not a "good" solution either (see the commit header).

I am not proud to suggest a "hacky" global setter function. You can guess that because in the commit header, I explain why I have put aside two "more clean" changes as too intrusive.

Would you be so kind as to explain what solution you have in mind?

Thanks.

asarhaddon:
Ping?

garyscavone:
How big does it need to be? Why not just make it bigger by default?

If that doesn’t work, then it would seem better to have an optional parameter in the constructor. Or we could consider creating a RtMidiIn::PortOptions structure as a new constructor argument that can be used in the future for things like this which may be API specific (similar to the StreamOptions in RtAudio).

keinstein:
I doubt that any particular size would match all cases. SysEx messages can have any arbitrarily large size.

At least also JACK uses a configurable fixed buffer size.

So the input and output buffers must be reconfigured when medium to large SysEx messages should be transferred.

I'd suggest to have two parameters: The largest message size and the minimum count of such messages that should fit into the buffers. These should be independent from the backend. Every backend can choose to calculate the right values itself or throw an exception if the requirement cannot be met.

So I'd prefer a generic way of configuration. This could also include the queue size parameter to the constructor.

asarhaddon:
Hello.
Currently, if an user requires a non-default buffer size, they have to edit `rtmidi.cpp`. Then, *on each rtmidi release*, they have to deal with a merge, rebuild both `rtmidi.o` and their project, then link statically.
With my suggestion, they can link with a precompiled librtmidi, either static or shared, like they would for most libraries. On each rtmidi update, they only need to rebuild their project. When my function is replaced with a proper solution, the user will have to update their project *only once*.
You are designing a more consistant interface. If someone plans to implement it before next release, I will happily close this merge request.  Else, my suggestion would improve the intermediate releases.
 I do not volunteer to design or implement the proper solution because I miss knowledge of audio systems and C++ skills.

asarhaddon:
Hello.
Apparently #264 implements a correct fix.
Thanks.

PR #190, opened 2019-02-23
Changes to CMakeLists to make use with add_subdirectory easier

mcclure:
There's a common technique with CMake where instead of building and installing your dependencies, you check your dependencies into your project as git submodules and then use add_subdirectory to make them build at the same time your project builds. This is good because it gets you reproducible builds and your versioning is enforced by git.

Right now RTMIDI does not work with add_subdirectory because it has variables/targets that are likely to namespace collide with other CMakeLists. For example right now if you use add_subdirectory to include RTMIDI and PhysFS in the same build it cannot build because they both declare a target named "uninstall" and CMake bails out when it sees this.

This patch fixes this by:

- "Namespace" option variables with RTMIDI_
- New RTMIDI_TARGETNAME_UNINSTALL option to optionally rename uninstall target

**Another thing that would help but is not in this patch** is if the source were moved into a src/ directory or the sources and includes moved into src/ and include/ directory. Right now when doing the add_subdirectory thing you have to include_directories() the whole RTMIDI project folder and this means the include path gets polluted with, like, your tests directory and stuff.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,10 +22,11 @@ if(WIN32)
 endif()
 
 # Build Options
-option(BUILD_SHARED_LIBS "Compile library shared lib." TRUE)
-option(BUILD_STATIC_LIBS "Compile library static lib." TRUE)
-option(BUILD_TESTING "Compile test programs." TRUE)
+option(RTMIDI_BUILD_SHARED_LIBS "Compile library shared lib." TRUE)
+option(RTMIDI_BUILD_STATIC_LIBS "Compile library static lib." TRUE)
+option(RTMIDI_BUILD_TESTING "Compile test programs." TRUE)
 set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Build type (Release,Debug)")
+set(RTMIDI_TARGETNAME_UNINSTALL "uninstall" CACHE STRING "Name of 'uninstall' build target")
 
 # API Options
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -206,7 +207,7 @@ message(STATUS "Compiling with support for: ${apilist}")
 # PkgConfig file
 string(REPLACE ";" " " req "${PKGCONFIG_REQUIRES}")
 string(REPLACE ";" " " api "${API_DEFS}")
-configure_file("rtmidi.pc.in" "rtmidi.pc" @ONLY)
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/rtmidi.pc.in" "rtmidi.pc" @ONLY)
```
mcclure: Note, I just realized after merging: The change on this line was accidental, but it is harmless I think. Sorry

radarsat1: I'm not sure if it's necessary but I saw it and it seems correct.  I suppose the 'current directory' is changed when descending directories, hence your comment?  In any case, harmless, as you say.  I'll leave it.

mcclure: I believe when you do the add_subdirectory, path roots (including both relatively specified files like this one, and the CMAKE_CURRENT_SOURCE_DIR variable) reset to the new subdirectory within the scope of that file. So this line change should have literally no effect I think

PR #192, opened 2019-03-16
Implement sendMessages

rianhunter:
Some backends may have an lower-latency way to send out multiple messages at once rather than one at time. This API allows clients to take advantage of that capability.

The first implementation is for `MidiOutAlsa`, which can buffer multiple messages at once before draining them out over the alsa pipe.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -2293,7 +2293,7 @@ void MidiOutAlsa :: openVirtualPort( const std::string &portName )
 
 void MidiOutAlsa :: sendMessage( const unsigned char *message, size_t size )
 {
-  int result;
+  long result;
   AlsaMidiData *data = static_cast<AlsaMidiData *> (apiData_);
   unsigned int nBytes = static_cast<unsigned int> (size);
   if ( nBytes > data->bufferSize ) {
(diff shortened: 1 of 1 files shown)


rianhunter:
Any chance of something like this getting merged?

rianhunter:
Gentle ping

rianhunter:
Any comments on this sort of change?

garyscavone:
I looked at this and tried it out myself. I have several reservations, including:

1. Need: Is efficiency and speed really an issue with MIDI? Personally, I found it less convenient in my tests to use the IOVec structure, rather than just send out a sequence of MIDI messages using the same std::vector<>.
2. The solution actually makes the default sendMessage() functions less efficient because all messages must be repackaged into an IOVec struct before output.
3. IOVec name: not very clear, as this is always used for 'O'utput, not input. Plus, it would be better if it has a name related to MIDI or messages.
4. The IOVec structure would need to be documented in doxygen

Another possible solution would be to use the current message variable types but change the parsing code to process multiple messages automatically. That said, it would make the sendMessage() function much more complex because of the added parsing.

Personally, I'm not convinced that this extra functionality is worth the extra baggage. If others feel strongly one way or the other, please chime in.

rianhunter:
> Need: Is efficiency and speed really an issue with MIDI? Personally, I found it less convenient in my tests to use the IOVec structure, rather than just send out a sequence of MIDI messages using the same std::vector<>.

At a high level the problem this is trying to solve is limiting system calls. System calls usually have a latency an order of magnitude or two higher than in-process operations. This change evolved from work I'm doing on a large scale real time system, where noticeable (and ugly sounding delays) occurred when the midi controller needed to send multiple simultaneous midi events (think multiple chords over sub 1/16th note rhythms).

> The solution actually makes the default sendMessage() functions less efficient because all messages must be repackaged into an IOVec struct before output.

In the default sendMessage implementation, the IOVec struct is allocated on the stack and passed by reference to the sendMessages function. That is far far less expensive in terms of latency than invoking the system call to send the midi messages to the operating system. That overhead is so negligible it's not even worth considering, it constitutes on the order of a dozen extra instructions.

Note that that overhead only happens when a native sendMessages is implemented, e.g. Linux. This change has no new performance effect on, say, macOS.

> IOVec name: not very clear, as this is always used for 'O'utput, not input. Plus, it would be better if it has a name related to MIDI or messages.
> The IOVec structure would need to be documented in doxygen

Yep.

garyscavone:
Thanks for the clarification, especially regarding the overhead only happening when native sendMessages is implemented ... I missed that. I'm still surprised you would be running into latency issues, as I've played back very complicated MIDI files using RtMidi and not noticed latencies in the past. That said, I don't do that much MIDI programming these days.

My only remaining questions would be:
1. I wonder if some APIs might inherently allow you to pass multiple MIDI messages in a single byte stream (I have a vague recollection of someone mentioning that for CoreMidi in the past)? If so, then the proposed IOVec structure might be unnecessary? 

2. I would still prefer a name for the structure that is clearer. MidiMessage is a little on the long side but clearer. There is already an internal MidiMessage structure in MidiInApi but since it is internal only, we could potentially rename it and steal the name. Or perhaps others might contribute other options.

3. Finally, I'm trying to wrap up changes (the low-hanging fruit) to make a new release very soon. The motivation is so that we can start a new branch that involves more significant changes (removing exceptions, better port identification, updating to more modern C++ functionality, etc..). In that case, perhaps this PR is better left to the new branch, especially since it involves an API change?

rianhunter:
> I wonder if some APIs might inherently allow you to pass multiple MIDI messages in a single byte stream (I have a vague recollection of someone mentioning that for CoreMidi in the past)? If so, then the proposed IOVec structure might be unnecessary?

That would be fine with me, though it might not be the most general API. I'm not sure if `snd_midi_event_encode()` is smart enough to only parse out the first message, maybe it is.

> I would still prefer a name for the structure that is clearer. MidiMessage is a little on the long side but clearer. There is already an internal MidiMessage structure in MidiInApi but since it is internal only, we could potentially rename it and steal the name. Or perhaps others might contribute other options.

If a separate struct is necessary, any name with which you are okay is okay with me.

> Finally, I'm trying to wrap up changes (the low-hanging fruit) to make a new release very soon. The motivation is so that we can start a new branch that involves more significant changes (removing exceptions, better port identification, updating to more modern C++ functionality, etc..). In that case, perhaps this PR is better left to the new branch, especially since it involves an API change

That would be fine and I defer to you. I would only note that this change is a backwards-compatible change, meaning it doesn't change previous behavior, it only adds new behavior, so it doesn't necessarily needed a major version bump.

garyscavone:
I verified that the current version of sendMessage() in MidiOutCore does support multiple messages if you comment out the check at line 1408. This should be checked with the other APIs but I won't have access to Windows and Linux for a few days.

rianhunter:
> I verified that the current version of sendMessage() in MidiOutCore does support multiple messages if you comment out the check at line 1408. This should be checked with the other APIs but I won't have access to Windows and Linux for a few days.

Very cool, I can check Linux later today. If that's the case, I'll format into a new smaller diff

garyscavone:
From a documentation check for ALSA and DS, the current functions used by RtMidi do not support multiple messages. I didn't check the Jack documentation but the function used seems so generic that  I expect it might support multiple messages. For ALSA, there may be other functions that support multiple messages?

rianhunter:
For ALSA, I think you may be able to use `snd_midi_event_encode` to parse out a byte stream of MIDI messages one by one. It returns the number of bytes consumed to make the message.

rianhunter:
@garyscavone Great suggestion, I've modified `MidiOutAlsa::sendMessage()` to accept multiple MIDI messages in a single buffer and it works beautifully, and requires no API changes. This capability can be added to other platforms as necessary but for the time being can be considered a platform-specific "easter egg" while full support is added to each platform.

rianhunter:
Gentle ping

rianhunter:
Should I close this PR?

PR #199, opened 2019-04-24
Only include CTest when buliding tests.

dpacbach:
This is because merely including CTest will actually enable
testing by calling CMakes enable_testing() function.

Without this fix, a parent project that includes RtMidi as a submodule
will have testing enabled which may cause undesirable effects in
other submodules.


dpacbach:
Ping!

PR #200, opened 2019-04-24
Make include folder public.

dpacbach:
This way, when a parent project depends on the `rtmidi` target
(e.g. in a git submodule configuration) it will be able to see
its header file.


radarsat1:
closing as this seems to have been superseded by #206

PR #206, opened 2019-07-31
Expose include path for RtMidi.h in target rtmidi

lubkoll:
Currently I need to manually add the include-path for RtMidi.h when using RtMidi. With this pull-request this will not be necessary any more and 

find_package(RtMidi REQUIRED)
target_link_libraries(my-target PRIVATE RtMidi::rtmidi)

will be all that is required to use your library, which is more elegant and less error-prone than manually configuring the include path. Hope you like it:)


liarokapisv:
This is very important for modern cmake integration.

liarokapisv:
wow, this was fast!

PR #212, opened 2020-02-27
Add Web MIDI API support with emscripten.

atsushieno:
Hello,

I have been trying to bring [JUCE](https://github.com/WeAreROLI/JUCE) audio plugins to the web using [Dreamtonics/juce_emscripten](https://github.com/Dreamtonics/juce_emscripten/).

When I thought of MIDI support there, I though that adding Web MIDI support to RtMidi would make it nicer as other people would benefit wasm support too.

Therefore, this changeset brings in Web MIDI API support.

It is enabled when  `__WEB_MIDI_API__` is defined. I'm not sure what is good for ordinal(?) emscripten environment so I left other build script parts untouched.

The changes are already used in https://github.com/Dreamtonics/juce_emscripten/pull/4.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -57,7 +57,7 @@
 //
 // **************************************************************** //
 
-#if !defined(__LINUX_ALSA__) && !defined(__UNIX_JACK__) && !defined(__MACOSX_CORE__) && !defined(__WINDOWS_MM__)
+#if !defined(__LINUX_ALSA__) && !defined(__UNIX_JACK__) && !defined(__MACOSX_CORE__) && !defined(__WINDOWS_MM__) && !defined(__WEB_MIDI_API__)
   #define __RTMIDI_DUMMY__
 #endif
 
@@ -231,6 +231,57 @@ class MidiOutWinMM: public MidiOutApi
 
 #endif
 
+#if defined(__WEB_MIDI_API__)
+
+class MidiInWeb : public MidiInApi
+{
+  std::string client_name{};
+  std::string web_midi_id{};
+  int open_port_number{-1};
+
+ public:
+  MidiInWeb(const std::string &/*clientName*/, unsigned int queueSizeLimit );
+  ~MidiInWeb( void );
+  RtMidi::Api getCurrentApi( void ) { return RtMidi::WEB_MIDI_API; };
+  void openPort( unsigned int portNumber, const std::string &portName );
+  void openVirtualPort( const std::string &portName );
+  void closePort( void );
+  void setClientName( const std::string &clientName );
+  void setPortName( const std::string &portName );
+  unsigned int getPortCount( void );
+  std::string getPortName( unsigned int portNumber );
+
+  void onMidiMessage( uint8_t* data, double domHishResTimeStamp );
+
+ protected:
+  void initialize( const std::string& clientName );
+};
+
+class MidiOutWeb: public MidiOutApi
+{
+  std::string client_name{};
+  std::string web_midi_id{};
+  int open_port_number{-1};
+
+ public:
+  MidiOutWeb( const std::string &clientName );
+  ~MidiOutWeb( void );
+  RtMidi::Api getCurrentApi( void ) { return RtMidi::WEB_MIDI_API; };
+  void openPort( unsigned int portNumber, const std::string &portName );
+  void openVirtualPort( const std::string &portName );
+  void closePort( void );
+  void setClientName( const std::string &clientName );
+  void setPortName( const std::string &portName );
+  unsigned int getPortCount( void );
+  std::string getPortName( unsigned int portNumber );
+  void sendMessage( const unsigned char *message, size_t size );
+
+ protected:
+  void initialize( const std::string& clientName );
+};
+
+#endif
+
 #if defined(__RTMIDI_DUMMY__)
 
 class MidiInDummy: public MidiInApi
@@ -299,6 +350,7 @@ const char* rtmidi_api_names[][2] = {
   { "alsa"        , "ALSA" },
   { "jack"        , "Jack" },
   { "winmm"       , "Windows MultiMedia" },
+  { "web"         , "Web MIDI API" },
   { "dummy"       , "Dummy" },
 };
 const unsigned int rtmidi_num_api_names =
@@ -318,6 +370,9 @@ extern "C" const RtMidi::Api rtmidi_compiled_apis[] = {
 #endif
 #if defined(__WINDOWS_MM__)
   RtMidi::WINDOWS_MM,
+#endif
+#if defined(__WEB_MIDI_API__)
+  RtMidi::WEB_MIDI_API,
 #endif
 #if defined(__RTMIDI_DUMMY__)
   RtMidi::RTMIDI_DUMMY,
@@ -401,6 +456,10 @@ void RtMidiIn :: openMidiApi( RtMidi::Api api, const std::string &clientName, un
   if ( api == MACOSX_CORE )
     rtapi_ = new MidiInCore( clientName, queueSizeLimit );
 #endif
+#if defined(__WEB_MIDI_API__)
+    if ( api == WEB_MIDI_API )
+    rtapi_ = new MidiInWeb( clientName, queueSizeLimit );
+#endif
 #if defined(__RTMIDI_DUMMY__)
   if ( api == RTMIDI_DUMMY )
     rtapi_ = new MidiInDummy( clientName, queueSizeLimit );
@@ -468,6 +527,10 @@ void RtMidiOut :: openMidiApi( RtMidi::Api api, const std::string &clientName )
 #if defined(__MACOSX_CORE__)
   if ( api == MACOSX_CORE )
     rtapi_ = new MidiOutCore( clientName );
+#endif
+#if defined(__WEB_MIDI_API__)
+    if ( api == WEB_MIDI_API )
+    rtapi_ = new MidiOutWeb( clientName );
 #endif
 #if defined(__RTMIDI_DUMMY__)
   if ( api == RTMIDI_DUMMY )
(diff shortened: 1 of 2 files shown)


radarsat1:
Cool!  I took the trouble to install emscripten in Docker and configured CMake to use `em++`, and it actually produced a WebAssembly module!

```
root@83abb940701a:~/rtmidi/build# file librtmidi.so.5.0.0 
librtmidi.so.5.0.0: WebAssembly (wasm) binary module version 0x1 (MVP)
root@83abb940701a:~/rtmidi/build# em++ --version
emcc (Emscripten gcc/clang-like replacement) 1.39.8 (commit 1458145cf4f3db0fb548343e6acab267eef8e4ef)
```

However, when I enabled `__WEB_MIDI_API__`, I got some errors:

```
root@83abb940701a:~/rtmidi/build# make
[  5%] Building CXX object CMakeFiles/rtmidi.dir/RtMidi.cpp.o
/root/rtmidi/RtMidi.cpp:238:15: error: function definition does not declare parameters
  std::string client_name{};
              ^
/root/rtmidi/RtMidi.cpp:239:15: error: function definition does not declare parameters
  std::string web_midi_id{};
              ^
/root/rtmidi/RtMidi.cpp:240:7: error: function definition does not declare parameters
  int open_port_number{-1};
      ^
/root/rtmidi/RtMidi.cpp:262:15: error: function definition does not declare parameters
  std::string client_name{};
              ^
/root/rtmidi/RtMidi.cpp:263:15: error: function definition does not declare parameters
  std::string web_midi_id{};
              ^
/root/rtmidi/RtMidi.cpp:264:7: error: function definition does not declare parameters
  int open_port_number{-1};
      ^
/root/rtmidi/RtMidi.cpp:3532:40: error: expected ';' after top level declarator
std::unique_ptr<WebMidiAccessShim> shim{nullptr};
                                       ^
                                       ;
/root/rtmidi/RtMidi.cpp:3545:34: error: expected expression
  return MAIN_THREAD_EM_ASM_INT( {
                                 ^
/root/rtmidi/RtMidi.cpp:3564:29: error: expected expression
  MAIN_THREAD_ASYNC_EM_ASM( {
                            ^
/root/rtmidi/RtMidi.cpp:3606:47: error: expected expression
  char *ret = (char*) MAIN_THREAD_EM_ASM_INT( {
                                              ^
/root/rtmidi/RtMidi.cpp:3638:17: error: variable has incomplete type 'void'
extern "C" void EMSCRIPTEN_KEEPALIVE rtmidi_onMidiMessageProc( MidiInApi::RtMidiInData* data, uint8_t* inputBytes, int32_t length, double domHighResTimeStamp )
                ^
/root/rtmidi/RtMidi.cpp:3638:37: error: expected ';' after top level declarator
extern "C" void EMSCRIPTEN_KEEPALIVE rtmidi_onMidiMessageProc( MidiInApi::RtMidiInData* data, uint8_t* inputBytes, int32_t length, double domHighResTimeStamp )
                                    ^
                                    ;
12 errors generated.
```

Maybe I am doing something wrong, I followed installation instructions [here](https://emscripten.org/docs/building_from_source/verify_emscripten_environment.html#verifying-the-emscripten-environment) and hacked the `CMakeLists.txt` a bit to define the flag.  How are you compiling/testing?

atsushieno:
Thanks for quick tryout!

What if you run `emmake make` instead of `make` ?

atsushieno:
I have been building and running code with juce_emscripten (linked in the first comment) which has its own build system (but it only #include-s the entire `RtMidi.cpp` from another `.cpp` source).

radarsat1:
I used emconfigure and emmake this time, but still got errors.  In fact RtMidi is not (yet) a C++11 codebase so we don't use the curly bracket initializers.  I got as far as removing them from the member variables, but got stuck on what to do about the "EM" macros:

```
make[1]: Entering directory '/root/rtmidi/build'
  CXX      librtmidi_la-RtMidi.lo
../RtMidi.cpp:3545:34: error: expected expression
  return MAIN_THREAD_EM_ASM_INT( {
                                 ^
../RtMidi.cpp:3564:29: error: expected expression
  MAIN_THREAD_ASYNC_EM_ASM( {
                            ^
../RtMidi.cpp:3606:47: error: expected expression
  char *ret = (char*) MAIN_THREAD_EM_ASM_INT( {
                                              ^
../RtMidi.cpp:3638:17: error: variable has incomplete type 'void'
extern "C" void EMSCRIPTEN_KEEPALIVE rtmidi_onMidiMessageProc( MidiInApi::RtMidiInData* data, uint8_t* inputBytes, int32_t length, double domHighResTimeStamp )
                ^
../RtMidi.cpp:3638:37: error: expected ';' after top level declarator
extern "C" void EMSCRIPTEN_KEEPALIVE rtmidi_onMidiMessageProc( MidiInApi::RtMidiInData* data, uint8_t* inputBytes, int32_t length, double domHighResTimeStamp )
```

atsushieno:
Alright - yeah I thought that some of those errors looked like from older C++ language version (and expected that using the right toolset may solve the issue, which didn't happen). I have no idea if it is possible to simply change the language version depending on the target (i.e. c++11 for web-midi), or whether it does make sense to have the code untested... Another choice would be to try to seek for another kind of coding that does not use `*_EM_ASM({})` syntax - it would be technically possible to pass JS string in C++ code, but the resulting code just becomes unmaintainable after a bunch of escaping) - the project policy is at your hand :-)

In the meantime, I have no problem to keep my wasm support code in my fork until you guys move forward at some stage to use C++11 and take the changes and/or poke me for further change requests, either closing or keeping this issue open.

radarsat1:
So emscripten is only usable from C++11 onwards?  (Just to clarify..)

atsushieno:
I only have like one week experience with emscripten, but to my understanding no - writing JS code in C++ will be just messy without [`*_EM_ASM({...})` macros](https://emscripten.org/docs/api_reference/emscripten.h.html) that makes code ugly (it would be with [`emscripten_run_script(" JS_as_string_escaped_literals ")`](https://emscripten.org/docs/api_reference/emscripten.h.html#c.emscripten_run_script).

atsushieno:
From https://emscripten.org/docs/api_reference/emscripten.h.html#c.EM_ASM :

> The C preprocessor does not have an understanding of JavaScript tokens, and as a result, if the code block contains a comma character ,, it may be necessary to wrap the code block inside parentheses. For example, code EM_ASM(return [1,2,3].length); will not compile, but EM_ASM((return [1,2,3].length)); does.

subsoap:
We're trying to get this working as a Defold extension but nothing we do gets Web/HTML5 version to work. Is there anything we're missing that would be obvious? @atsushieno 

One thing I had to do is remove try/catch blocks to get the emscripten version to compile.

It compiles and inits but can't open sockets. Other platforms work.

We do have ```__WEB_MIDI_API__``` defined, looked in the source to see what was needed. It needs to be added to documentation still / there might be other things missing in the documentation for using the web version? Asking here about stuff because no docs / not sure if anyone else has got the web version working.

https://github.com/subsoap/defold-midi/tree/dev-britzl

atsushieno:
Your project does not seem like a normal C/C++ project that I can easily guess what's going on there (I would not dive into your project's build system), but there is no socket dependency in my code. Maybe you are enabling some unnecessary platform build flags as well, e.g. for Linux (as Emscripten builds would be closer to it).

If your build system does not support C++ try-catch blocks then skip `rtmidi_c.cpp` which is AFAIK the only code file that had to make use of them. I didn't use it when I used my wasm driver at Dreamtonics/juce_emscripten.

britzl:
> but there is no socket dependency in my code

> It compiles and inits but can't open sockets

@atsushieno what @subsoap probably meant to write was that we can open a port using `openPort(number, name)`. I'm testing with https://flit.github.io/projects/midikeys/ on macOS and can open a port and get messages from MidiKeys.

When doing the same using the web build in Chrome (and granting the page permission) I am able to open a port but I get nothing in `getMessage(message)`.

Although I am now testing with `setCallback()` which is in fact working! So I guess there's a problem with `getMessage()` only then!

Another thing I noticed is that `isPortOpen()` always returns false in a web build. Wouldn't it make sense to set `connected_ = true` here: https://github.com/thestk/rtmidi/blob/master/RtMidi.cpp#L3784 and back to false again in `closePort`?

atsushieno:
> When doing the same using the web build in Chrome (and granting the page permission) I am able to open a port but I get nothing in getMessage(message).

Ah, that would explain it. Though I don't think that blocking (synchronous) requests for incoming message would work on Javascript on the web browsers. That's rather a platform limitation that is common to many other technologies.

> Another thing I noticed is that isPortOpen() always returns false in a web build. Wouldn't it make sense to set connected_ = true here: https://github.com/thestk/rtmidi/blob/master/RtMidi.cpp#L3784 and back to false again in closePort?

That's a good catch. There is no "connection state" in MIDI API and I never needed it, but probably it makes better sense for consistency across platforms. Maybe you can create a PR for the change? I'm just a one-time contributor and not at a position to approve your suggestion.

britzl:
> Ah, that would explain it. Though I don't think that blocking (synchronous) requests for incoming message would work on Javascript on the web browsers. That's rather a platform limitation that is common to many other technologies.

You're right. I was thinking that if no callback is set then messages could be queued and retrieved using `getMessage()`

> That's a good catch. There is no "connection state" in MIDI API and I never needed it, but probably it makes better sense for consistency across platforms. Maybe you can create a PR for the change?

Sure, will do!

britzl:
@atsushieno another question:

I noticed that the message bytes are never cleared after invoking the callback. Is this intentional?

https://github.com/thestk/rtmidi/blob/master/RtMidi.cpp#L3756-L3763

Most (but not all?) of the other platforms do a `message.bytes.clear()`

* https://github.com/thestk/rtmidi/blob/master/RtMidi.cpp#L926
* https://github.com/thestk/rtmidi/blob/master/RtMidi.cpp#L985
* https://github.com/thestk/rtmidi/blob/master/RtMidi.cpp#L2654

And so on...

atsushieno:
> I noticed that the message bytes are never cleared after invoking the callback. Is this intentional?

not what I am aware of && I remember, seems like another issue.

PR #219, opened 2020-05-23
macOS multithreading fix/note

tschiemer:
In multithreaded environments on macOS port enumeration etc might not work properly due to the system run loops not being called.

This is not directly an RtMidi bug, but for a simple to use library this might be (at least) mentioned and possibly an equally simple solution presented.

I don't like introducing system specific functions (`RtMidi_multithreadRunLoop()`), but this is the most simple approach as the individual API implementations are not exposed, which I think should be done..

> Note that notifyProc will always be called on the run loop which was current when MIDIClientCreate was first called.

https://developer.apple.com/documentation/coremidi/1495360-midiclientcreate?language=objc
https://ios.developreference.com/article/22075958/MidiClientCreate+notifyproc+not+getting+called

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,11 +205,13 @@ if (RTMIDI_BUILD_TESTING)
   add_executable(qmidiin    tests/qmidiin.cpp)
   add_executable(sysextest  tests/sysextest.cpp)
   add_executable(apinames   tests/apinames.cpp)
+  add_executable(backgroundscanner   tests/backgroundscanner.cpp)
   list(GET LIB_TARGETS 0 LIBRTMIDI)
-  set_target_properties(cmidiin midiclock midiout midiprobe qmidiin sysextest apinames
+  set_target_properties(cmidiin midiclock midiout midiprobe qmidiin sysextest apinames backgroundscanner
     PROPERTIES RUNTIME_OUTPUT_DIRECTORY tests
                INCLUDE_DIRECTORIES ${CMAKE_CURRENT_SOURCE_DIR}
                LINK_LIBRARIES ${LIBRTMIDI})
+  target_compile_options(backgroundscanner PUBLIC "-std=c++11")
   add_test(NAME apinames COMMAND apinames)
 endif()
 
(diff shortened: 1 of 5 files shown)


tschiemer:
Although I'm not 100% certain (not too deep an insight here), I would think it would have to be run recurringly to process events, otherwise they'd not be delivered to the RtMidi subscriber - so a single call in the constructor would not lead to the desired result.

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/RunLoopManagement/RunLoopManagement.html
CFRunLoopRunInMode(https://developer.apple.com/documentation/corefoundation/1541988-cfrunloopruninmode?language=objc

But please feel free to test ;)

radarsat1:
Oh I see, it's more like a polling function, I didn't read your example properly.  But, from your example, it seems like something really specific to Mac (sorry I don't have one for testing), and you are literally checking if you are on OS X and calling a special function:

```
 while(1){
    #ifdef __APPLE__
    RtMidi_multithreadRunLoop();
    #endif
  }
```

But if that's the case, if this is something you know you have to do only on OS X, and you have to go out of your way like this to do it, I don't see why it's RtMidi's responsibility?  You could just as easily do,

```
 while(1){
    #ifdef __APPLE__
    CFRunLoopRunInMode( kCFRunLoopDefaultMode, 0, false );
    #endif
  }
```

I don't really see what wrapping this call really adds, as it doesn't provide any OS transparency, which is the whole point of RtMidi.

On the other hand if RtMidi needed a general polling function for the main loop,

```
 while(1){
    RtMidi::pollMessages();
  }
```

it might make some sense, but then I'm not sure of the benefit of using threading here.

garyscavone:
I generally agree with @radarsat1 on this. If this is still an issue, I would say that it would be nice to add some information about it in the doxygen documentation (probably in API Notes section?). This PR has conflicts so it would need to be resolved and updated.

insolace:
Thanks for this, @tschiemer, and apologies it sat so long.

@radarsat1's objection stands: the run loop has to be pumped repeatedly, so the most RtMidi could offer is a polling function the caller invokes in their own loop behind `__APPLE__`. That is barely better than calling `CFRunLoopRunInMode()` directly.

Gary suggested a doxygen note instead, prefaced with "if this is still an issue." We went looking. The crash behind this, #262, was closed yesterday by its own reporter: "I haven't seen this issue in years, maybe it was a MacOS bug and Apple fixed it." We had also tried to reproduce it on macOS 26.4.1, both same-thread and cross-thread, four runs, no crash either way.

So there is nothing current to document against, and a note describing a constraint we cannot demonstrate would mislead more than it helps. Closing with no code change and no note. If anyone hits this on a current macOS, open an issue with a repro and we will revisit.

Issue #155, opened 2018-02-17
Intermittent initialization issues on macOS High Sierra


jceipek:
I'm using RtMidi by calling into a .bundle I generated from the latest version of the included C wrapper without any modifications. It mostly works fine, but intermittently I get a `MidiInCore::initialize: error creating OS-X MIDI client object (-50)` error and then I won't be able to use `rtmidi_in_create_default` or `rtmidi_out_create_default` anymore (they fail with the same error) until I forcibly unload the .bundle. Even reconnecting my MIDI device doesn't fix the problem.

Prior to using macOS High Sierra, I wasn't encountering any issues. Note that I am calling `rtmidi_out_free` and `rtmidi_in_free` after closing every port I open.

Do you have any suggestions for tracking down the problem?
I'm at a loss for how to debug this, because the initialization seems to go fine until https://github.com/thestk/rtmidi/blob/master/RtMidi.cpp#L626, where `MIDIClientCreate` returns `-50`, an error code that doesn't seem to be defined in `MIDIServices.h` and seems to typically mean "Invalid Parameter".

Could the previous client be sticking around somehow, and if so, how could I determine that and release it if necessary?

radarsat1:
Hm, I'm not sure how to go about it exactly.. I guess what I would do in this case is cut out the CoreMidi part of the code and produce a minimal example using only CoreMidi that demonstrates the problem, and go to a Mac developer forum.  Especially if it's returning undefined errors, it could be an operating system/driver bug that needs to be reported to Apple.  (That's just a shot in the dark of course..)

jceipek:
I just found this post by someone with a similar issue but way more experience with C++ and CoreMidi than me:

https://lists.apple.com/archives/coreaudio-api/2013/Sep/msg00014.html

From the post:
"""
ret = MIDIClientCreate(cname, NULL, NULL, &mclient);

It gives me ret == -50 , which is a (paramErr): error in user parameter list
"""

and 

"""
I've resolved this error, it's ok. It was caused by an uninitalised variable.
"""

Could that be what's happening here?

jceipek:
Also: https://stackoverflow.com/questions/29411341/midiclientcreate-the-operation-couldnt-be-completed-osstatus-error-50

> MIDIClientCreate itself will return -50 in only two scenarios:
> 1) If outClient itself is NULL so in the above you forget to take the address of client.
> 2) If the MIDI server has an issue getting a process ID back internally

radarsat1:
[...]

PR #220, opened 2020-05-23
MIDIClientRef singleton (macOS)

tschiemer:
Addresses  https://github.com/thestk/rtmidi/issues/155#issuecomment-626181015

Not thoroughly tested yet but so far no problems in my multithreaded development project with regular instantiation and deletion of RtMidi-Objects..

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,9 +32,9 @@ In some cases, for example to use RtMidi with GS Synth, it may be necessary for 
 
 ## OSX / macOS
 
-- An essential component of the macOS MIDI subsystem (`MIDIClient`) can exhibit buggy behaviour (also see https://github.com/thestk/rtmidi/issues/155). As the relevant API is not directly exposed, the two following functions are declared (if and only if `__APPLE__` is defined) if access to this part is needed (but typically is not): 
- - `void RtMidi_setCoreMidiClientSingleton(MIDIClientRef client)`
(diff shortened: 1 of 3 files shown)


Noir-:
Would you mind merging this? This is a breaking issue and the whole open source world relies upon this library :)

garyscavone:
No time to test … hopefully it doesn’t create a problem.

On Apr 21, 2021, at 2:59 PM, threa4ctor ***@***.******@***.***>> wrote:



Would you mind merging this? This is a breaking issue and the whole open source world relies upon this library :)

garyscavone:
I don't see the need for the singleton function declarations to be in RtMidi.h ... I just moved them to RtMidi.cpp and it seems to work. Please let me know ASAP if this is a problem, as I will make the change shortly for the new release.

tschiemer:
No, it is indeed not needed but was primarily ment to allow direct access to the system object, if so desired (or possibly needed, can't tell)

Please note the relevant readme passage which should be removed aswell.

Just *looking* at the code in `RtMidi.cpp` I would think you can also completely remove the declarations along with the implementations as I don't see them being used in the first place; the singleton is accessed directly in the respective MidiIn/OutCore methods.

If it works for you I would think it's safe to remove

garyscavone:
Thanks! Fixed in latest two PRs.

PR #227, opened 2020-10-15
Fix memory leak from rtmidi_get_port_name

torkeldanielsson:
The address sanitizer warned of a leak in our code linked to rtmidi_get_port_name. One fix could be to just let the recipient of the char * free it. But when compiled as a dll I think that can lead to strange behavior if the standard libraries used in dll and main application differ. So I propose changing the interface of the rtmidi_get_port_name function so the user of it provides a char buffer for it to write into.

### DIFF
--- a/rtmidi_c.cpp
+++ b/rtmidi_c.cpp
@@ -132,17 +132,27 @@ unsigned int rtmidi_get_port_count (RtMidiPtr device)
     }
 }
 
-const char* rtmidi_get_port_name (RtMidiPtr device, unsigned int portNumber)
+int rtmidi_get_port_name (RtMidiPtr device, unsigned int portNumber, char * bufOut, int * bufLen)
 {
-    try {
-        std::string name = ((RtMidi*) device->ptr)->getPortName (portNumber);
-        return strdup (name.c_str ());
+    if (bufOut == nullptr && bufLen == nullptr) {
+        return -1;
+    }
 
+    std::string name;
+    try {
+        name = ((RtMidi*) device->ptr)->getPortName (portNumber);
     } catch (const RtMidiError & err) {
         device->ok  = false;
         device->msg = err.what ();
-        return "";
+        return -1;
     }
+
+    if (bufOut == nullptr) {
+        *bufLen = static_cast<int>(name.size());
+        return 0;
+    }
+
+    return snprintf(bufOut, static_cast<size_t>(*bufLen), "%s", name.c_str());
 }
 
 /* RtMidiIn API */
--- a/rtmidi_c.cpp
+++ b/rtmidi_c.cpp
@@ -148,7 +148,7 @@ int rtmidi_get_port_name (RtMidiPtr device, unsigned int portNumber, char * bufO
     }
 
     if (bufOut == nullptr) {
-        *bufLen = static_cast<int>(name.size());
+        *bufLen = static_cast<int>(name.size()) + 1;
         return 0;
     }
 
--- a/rtmidi_c.h
+++ b/rtmidi_c.h
@@ -155,10 +155,14 @@ RTMIDIAPI void rtmidi_close_port (RtMidiPtr device);
  */
 RTMIDIAPI unsigned int rtmidi_get_port_count (RtMidiPtr device);
 
-/*! \brief Return a string identifier for the specified MIDI input port number.
+/*! \brief Access a string identifier for the specified MIDI input port number.
+ * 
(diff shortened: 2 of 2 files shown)


marcel303:
Actually this is not a memory leak, but (worse) it's returning a pointer to previously freed memory (std::string goes out of scope, and frees its memory), which may potentially lead to a crash. The proposed solution makes sense.

keryell:
The original code is correct with the `strdup` doing the heap allocation and copy, so independent of the end-of-life of the `std::string`. So the address sanitizer detect some lacking `free` in the calling code.
What is not correct is returning `""` when there is an error because this memory cannot be freed but the caller cannot really now.
So, either returning `strdup("")` or `nullptr`, by specifying in the API that `nullptr` is returned in case of error.
Or even better, not using an unsafe C API, but this is another story...

rtmidi_c.cpp
```diff
@@ -132,17 +132,27 @@ unsigned int rtmidi_get_port_count (RtMidiPtr device)
...
-        return "";
+        return -1;
     }
+
+    if (bufOut == nullptr) {
+        *bufLen = static_cast<int>(name.size());
+        return 0;
+    }
+
+    return snprintf(bufOut, static_cast<size_t>(*bufLen), "%s", name.c_str());
```
keryell, line 155: Otherwise there is https://en.cppreference.com/w/c/string/byte/strncpy

marcel303:
> The original code is correct with the `strdup` doing the heap allocation and copy, so independent of the end-of-life of the `std::string`. So the address sanitizer detect some lacking `free` in the calling code.
> What is not correct is returning `""` when there is an error because this memory cannot be freed but the caller cannot really now.
> So, either returning `strdup("")` or `nullptr`, by specifying in the API that `nullptr` is returned in case of error.
> Or even better, not using an unsafe C API, but this is another story...

You're correct. I failed to notice the strdup call there and it's implications..

Another possible solution (requiring an api change), is to change the function signature to
    int rtmidi_get_port_name (RtMidiPtr device, unsigned int portNumber, char ** bufOut)

char * name;
if (rtmidi_get_port_name(device, portNumber, &name)

I usually find this pattern quite helpful to understand a function will allocate the output buffer.

torkeldanielsson:
> The original code is correct with the `strdup` doing the heap allocation and copy, so independent of the end-of-life of the `std::string`. So the address sanitizer detect some lacking `free` in the calling code.

I disagree. This code ends up in a dll. A call to 'free' in the calling code is not guaranteed to end up in the same allocator. The principle is that you have to free memory with the same allocator that allocated it.

keinstein:
I don't see any difference between returning `char *` and using a `char **` argument. Both allow to to return a user managed pointer as well as a library managed pointer. 

@torkeldanielsson You are right. The C API should alias the free function as `rtmdi_free` in the DLL. Note: On *nix platforms this isn't usually a big problem as `libc` is dynamically linked to all libraries at once, while on windows each DLL is linked separately against `libc`. You can even link a DLL statically.

marcel303:
> I don't see any difference between returning `char *` and using a `char **` argument. Both allow to to return a user managed pointer as well as a library managed pointer.

Actually a decision needs to be made if the library should allocate the memory, or the user.

The change list mentioned by @torkeldanielsson lets the user first request the size of the buffer, allocate it themselves, and in a next call use it to return the name. So no rtmdi_free is needed. Letting the library allocate the buffer would be possible with the `char**` version, but then you'll need to call rtmdi_free.

I'm biasing towards the change made by @torkeldanielsson. In his code it's also ok to just use a buffer without requesting the size requirement first. So you could just use a fixed size buffer of say 256 bytes, assuming names won't exceed this length. snprintf will always null terminate the buffer, so even if the buffer is too small it won't be a bug issue.

@torkeldanielsson I think to make the api more usable, the returned length should include space for the null byte. So `name.size() + 1` instead of just `.size()`.

    if (bufOut == nullptr) {
         *bufLen = static_cast<int>(name.size());
         return 0;
     }

> 
> @torkeldanielsson You are right. The C API should alias the free function as `rtmdi_free` in the DLL. Note: On *nix platforms this isn't usually a big problem as `libc` is dynamically linked to all libraries at once, while on windows each DLL is linked separately against `libc`. You can even link a DLL statically.

torkeldanielsson:
> the returned length should include space for the null byte. So name.size() + 1 instead of just .size()

Good point! I changed it to be like that.

PR #228, opened 2020-12-05
Added option to MidiApi to stop printing errors and warnings

SarahVIAUD:
Added a static function to MidiApi to be able to stop printing warnings and errors to the console.
This is especially useful when RtMidi object is initialised with no device connected yet, we don't necessary want a warning being printed.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -533,6 +533,12 @@ RtMidiOut :: ~RtMidiOut() throw()
 //*********************************************************************//
 //  Common MidiApi Definitions
 //*********************************************************************//
+bool MidiApi :: printErrorIfNoCallback_ = false;
+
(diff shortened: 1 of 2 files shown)


keinstein:
Why didn't you make the printing itself a callback, which is by default initalised. This could use the current API. Exposing it in RtMidi.h would allow the user code to fall back to the default. The benefit is that you have to keep track of it in only one place.

SarahVIAUD:
"The printing itself a callback"? Not sure I understand what you mean...
I do setup the error callback but for that I first have to create the RtMidiIn instance and the creation generates a warning message by itself when no device available because the error callback function cannot have been initialised yet.
With the current API I do not see a way to create RtMidiIn with the callback function already set up.

PR #245, opened 2021-04-09
Allow defining the sysex buffer size.

justinlatimer:
On Windows, buffers are initialized and passed through the system APIs for use with sysex messages. This change allows applications to override the default buffer sizes.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -2423,8 +2423,12 @@ static std::string ConvertToUTF8(const TCHAR *str)
   return u8str;
 }
 
+#ifndef RT_SYSEX_BUFFER_SIZE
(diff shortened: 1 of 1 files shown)


justinlatimer:
As an alternative to an existing pull request (#186) this is the simplest possible change to allow Windows users a level of customisation of buffer sizes without changing the API of the library.

keinstein:
Some nitpicking: Your patch does not allow the application to calculate or ask the user for the SysEx size. It only allows the *developer* to define how large the SysEx can be.

justinlatimer:
That's fair — I should have said "Windows users of the library".

garyscavone:
As far as I can tell, this issue has been dealt with via the merged PR #264.

PR #259, opened 2021-08-16
locking for queue

folkertvanheusden:
The queue is accessed from a thread as well as main so requires locking.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -661,8 +661,9 @@ double MidiInApi :: getMessage( std::vector<unsigned char> *message )
   return timeStamp;
 }
 
+// structure should be locked before using this method
 unsigned int MidiInApi::MidiQueue::size( unsigned int *__back,
-                                         unsigned int *__front )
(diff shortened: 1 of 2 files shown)


radarsat1:
Did you observe any bad behaviour any specific platform that triggered work on this PR?

I am asking because as far as I know the message queue implements a unidirectional ring buffer that should not require locking.  (Push only updates back and pop only updates front.)  So if you've seen messages getting corrupted it might be a bug rather than that a lock is required.

keinstein:
When properly accessed a ring buffer is a FIFO which can be accessed by one thread in write-only mode (modify the data and advance the front pointer) and another thread in read-only mode (read the data between the back and the front pointer and advance the back pointer). Pointer arithmetics is done modulo the buffer length.

keinstein:
BTW if you really need locking, you should use spinlocks in that situation. They are much faster than normal mutexes at a cost of CPU time. But the Operations shouldn't be that expensive.

PR #260, opened 2021-08-16
Unplug

folkertvanheusden:
Throw an error back to the calling python script when the (virtual/physical-) MIDI device got unplugged.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -661,8 +661,9 @@ double MidiInApi :: getMessage( std::vector<unsigned char> *message )
   return timeStamp;
 }
 
+// structure should be locked before using this method
 unsigned int MidiInApi::MidiQueue::size( unsigned int *__back,
-                                         unsigned int *__front )
+                                         unsigned int *__front)
 {
   // Access back/front members exactly once and make stack copies for
   // size calculation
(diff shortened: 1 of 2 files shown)


insolace:
I'm not sure Windows provides any useful means of knowing when a device is unplugged. Our approach has been to store all of the MIDI port numbers/names in a table and then regularly polling them for changes. If there's a change, then we call an update function.

keinstein:
The queue is accessed as a ring buffer. That means the buffer is intended to be accessed by exactly one thread in parallel from the user code as well by the internal thread. The responsibilities are spread between the threads. The realtime thread writes data and increases the end pointer while the user thread increases the start pointer. Race conditions can occur when one pointer is close to the other but this means that he user thread reads beyond end or the realtime thread recognises a buffer overrun. So the worst thing that can happen is that the realtime thread is informed too late about free memory. This is a race condition that has no logical solution. 

Practically the compiler optimisations or the processor cache may cause trouble. In such a case a memory barrier must be installed in order to make sure that all data is written before it is read.

If you need several user  threads to access the queue you can add locking mechanisms there.

Furthermore, in realtime functions probably spinlocks are a better choice as they return faster.

sagamusix:
> I'm not sure Windows provides any useful means of knowing when a device is unplugged

Indeed there isn't. What I'm doing in my own application (but would be difficult to add to RtMidi as-is) is to listen to `WM_DEVICECHANGE` messages, and if the event type is `DBT_DEVNODES_CHANGED`, I call `midiInGetNumDevs()`, which is enough to provoke the MIDI driver to send a `MIM_CLOSE` message in the MIDI callback, which can then be acted upon.

Issue #275, opened 2021-11-21
JACK support broken on macOS


OPNA2608:
#### Expected behavior and actual behavior.

- Expected: JACK API support should compile on macOS, like it did on RtMidi 4.0.0.

- Actual: Since #203, this appears to be broken.
  ```
  [  5%] Building CXX object CMakeFiles/rtmidi.dir/RtMidi.cpp.o
  /tmp/nix-build-rtmidi-5.0.0.drv-0/source/RtMidi.cpp:3611:7: error: use of undeclared identifier 'pthread_yield'
        pthread_yield();
        ^
  1 error generated.
  make[2]: *** [CMakeFiles/rtmidi.dir/build.make:76: CMakeFiles/rtmidi.dir/RtMidi.cpp.o] Error 1
  make[1]: *** [CMakeFiles/Makefile2:126: CMakeFiles/rtmidi.dir/all] Error 2
  make: *** [Makefile:136: all] Error 2
  ```
  This function [doesn't seem to exist](https://opensource.apple.com/source/libpthread/libpthread-454.80.2/include/pthread/pthread.h.auto.html) on macOS. There's a `pthread_yield_np` one though.

#### Steps to reproduce the problem.

0. Get macOS
1. Get JACK on macOS (i.e. via homebrew)
2. Try to compile RtMidi with JACK support

#### Specifications like the version of the project, operating system, or hardware.

- RtMidi Version: 5.0.0
- Operating System: macOS 10.12 or higher (I'm unaware precisely what version our downstream builders are running)
- Hardware: x86_64 & apple aarch64

PR #278, opened 2021-12-30
Use posix sched_yield instead of pthread_yield

niclasr:
Use posix sched_yield instead of pthread_yield.
pthread_yield is linux specific sched_yield is
a standard posix function. pthread_yield on linux
is implemented using sched_yield.
This makes the jack plugin work on other
platforms than linux. Fixes build of Jack on NetBSD.


niclasr:
Ok pthread_yield is not linux specific but it is not available everywhere.

niclasr:
This pull request should fix #275

PR #288, opened 2022-03-18
Better linking of ALSA

dliess:
I ran into a problem when using rtmidi in a yocto-SDK due to the problem, that when ${ALSA_LIBRARY} is used for linking, then it links against the whole Path to the alsa so-file and that path is incorrect in an sdk environment. It is cleaner to link against the cmake-target ALSA:ALSA and add the missing "find_dependency" to the target.cmake file.


beldenfox:
The target ALSA::ALSA is defined in CMake version 3.12 and above. This change causes errors to be thrown on Ubuntu 18 LTS which uses CMake version 13.10.

PR #296, opened 2022-06-17
Keep compile flags private

observant2:
Hi! I used rtmidi with vcpkg and [ran into problems](https://github.com/microsoft/vcpkg/issues/25279) with compiling it with `CMAKE_BUILD_TYPE debug`. I'm not sure if this pull request fixes the problem, but if I understand it right then this at least doesn't propagate Werror to projects that add rtmidi as a dependency.

Also `CMAKE_COMPILER_IS_GNUCXX` is deprecated according to CMake docs, so I replaced it with what was written there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,19 +46,6 @@ endif()
 option(RTMIDI_API_WINMM "Compile with WINMM support." ${WIN32})
 option(RTMIDI_API_CORE "Compile with CoreMIDI support." ${APPLE})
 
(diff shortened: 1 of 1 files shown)


garyscavone:
Does anyone have any feedback on this PR? I don't use CMake and the author isn't so confident that this actually fixes a problem.

CasparKielwein:
The patch looks fine to me.

I can't judge if it fixes the issue of OP, but it is a correct change to make compile flag target private.

PR #311, opened 2023-02-10
Realign enum order across C and C++ interfaces

jbarrett:
It looks like the new Web API didn't make it to the header for the C wrapper. This change adds the value to the C API's enum, and changes the order in the C++ header to align with `rtmidi_api_names` here: https://github.com/thestk/rtmidi/blob/806e18f575b68c23b26f9398e1b6866b335b5308/RtMidi.cpp#L375-L386

(I think pushing the "dummy" interface to the end makes sense...)

Thanks!


ejconlon:
Hi John, I agree this needs to be done. However, I think we should stick with the tried and true strategy of only appending new values to the enum definition! This is a working patch I apply in the Haskell bindings: https://github.com/riottracker/RtMidi/blob/f006e1d040f05a7974910d2392c09090bf71a8bb/rtmidi.patch .

garyscavone:
I think this update was incorporated in another PR I merged recently ... it is no longer necessary.

PR #317, opened 2023-07-08
Added an Android Studio project

YellowLabrador:
A simple Android Studio project showing how to build RtMidi from the IDE. It includes a very simple app.


insolace:
You are a true hero

wildgoose789:
Apologies, a novice here looking for something for Unity in c# that this feels like it could help with - trying to get an oculus quest to recognise MIDI within a unity build - is there a way to use rtmidi for this now android has been added? if so how (for someone not familiar with c++ sorry!!)

PR #335, opened 2024-03-20
Fix a CFStringRef leak in RTMidi Port Name

baconpaul:
In certain circumstances when building a compound port name (for instance, with an IAC bus), the string math to concatenate the bus and driver does not release one of the source strings, causing a CFStringRef leak on macOS.

This fixes it by doing the appropriate CFRelease after string concatenation.


baconpaul:
Thanks! 

Curious if you have a cadence to tag changes. Seems like there’s a few changes since 66.0.0 so wondering if a 6.0.1 might be on the horizon?

appreciate the merge

PR #344, opened 2024-10-22
fix: link error for testcapi

franko:
When building the library as a static library it tries to link testcapi using the C compiler and we get a link error because the C++ support is missing.

We enable link of testcapi using C++ when shared library is disabled.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,6 +84,7 @@ AS_IF([test "x$AR" = "xno"], [
 ])
 
 LT_INIT([win32-dll])
+AM_CONDITIONAL([enable_shared], [test "$enable_shared" = "yes"])
 AC_CONFIG_MACRO_DIR([m4])
 
 # Checks for header files.
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -31,6 +31,9 @@ apinames_LDADD = $(top_builddir)/librtmidi.la
 
 testcapi_SOURCES = testcapi.c
 testcapi_LDADD = $(top_builddir)/librtmidi.la
+if !enable_shared
+testcapi_LINK = $(CXXLINK)
+endif
 
 EXTRA_DIST = cmidiin.dsp midiout.dsp midiprobe.dsp qmidiin.dsp	\
 	sysextest.dsp RtMidi.dsw


insolace:
Apologies this sat without a response for so long — and apologies for the truncated comment just above, which was my mistake. Full version below.

I picked this up while going through open build-system PRs. Short version: **the bug you found is real and still present**, your diagnosis is correct, but the patch as written breaks the default (shared) build. There is a smaller fix that works in both configurations — details below, and a question for you at the end.

## The bug reproduces

Upstream `a3233c2`, Autotools, `--disable-shared --enable-static`, Linux/ALSA:

```
librtmidi.a(librtmidi_la-RtMidi.o): undefined reference to `operator delete(void*)'
librtmidi.a(librtmidi_la-RtMidi.o): undefined reference to `std::locale::~locale()'
librtmidi.a(librtmidi_la-RtMidi.o): undefined reference to `vtable for __cxxabiv1::__si_class_type_info'
make[1]: *** [Makefile:666: testcapi] Error 1
```

413 undefined references in total. Exactly as you described: `testcapi.c` is linked with the C driver, and a static `librtmidi.a` does not pull in the C++ runtime.

Worth noting this is **Autotools-only** — CMake with `BUILD_SHARED_LIBS=OFF` builds and links `testcapi` cleanly, because it drives the link through the C++ toolchain already.

## Your patch fixes static, but regresses shared

With the patch applied:

| configuration | `make` | notes |
|---|---|---|
| `--disable-shared --enable-static` | **0** | fixed; `make check` 1 PASS, `testcapi` runs |
| default (shared) | **2** | regression |

The shared failure:

```
  GEN      testcapi
/bin/bash: line 1: testcapi.o: command not found
make[1]: *** [Makefile:667: testcapi] Error 127
```

Note `GEN` where an unpatched build shows `CCLD`. In the shared configuration the `if !enable_shared` branch is not taken, so `testcapi_LINK` is left empty, and Automake emits a link rule with no linker command — the bare object name is then executed as a program. Control run on the same machine without the patch: `make` 0, `make check` 1 PASS. So it is the patch, not the environment.

## A smaller fix that works in both

This needs no `configure.ac` change and no conditional, so there is no untested branch to get wrong:

```makefile
testcapi_SOURCES = testcapi.c
testcapi_LDADD = $(top_builddir)/librtmidi.la
# testcapi.c is C, but librtmidi is C++: linking it with the C driver
# fails against a static libtool archive, which does not pull in the C++
# runtime. Naming a C++ source here makes Automake choose the C++ linker.
# The file is never compiled and need not exist; only its suffix matters.
nodist_EXTRA_testcapi_SOURCES = dummy.cxx
```

`nodist_EXTRA_..._SOURCES` is the documented Automake idiom for forcing C++ linkage on a C program. Automake reads only the suffix when choosing the linker; `dummy.cxx` is never compiled and is not created on disk (verified).

Tested on Linux/ALSA, all four configurations:

| configuration | `make` | link | undefined refs | `make check` |
|---|---|---|---|---|
| Autotools static | 0 | `CXXLD` | 0 | 1 PASS / 0 FAIL |
| Autotools shared | 0 | `CXXLD` | 0 | 1 PASS / 0 FAIL |
| CMake static | 0 | — | 0 | `testcapi` runs |
| CMake shared | 0 | — | 0 | `testcapi` runs |

`testcapi` runs correctly in every case.

## Unrelated, but found while testing

`make dist` fails on current `master` with `No rule to make target 'cmidiin.dsp', needed by 'distdir'` — `EXTRA_DIST` in `tests/Makefile.am` lists MSVC `.dsp` files that are no longer in the tree. Pre-existing and independent of this PR (it fails identically on unpatched `master`); mentioning it only so it is not mistaken for fallout from either patch.

## How would you like to proceed?

Your call, and happy either way:

1. **You amend this PR** — with the one-line version above, or any other approach you prefer. It stays yours, which seems right since you found and diagnosed it. **If you submit it, we will merge it.**
2. **We open a PR carrying the fix** and credit you explicitly for the diagnosis and the original patch, leaving this one for the maintainer to close.

No rush on a reply. If we do not hear from you in a week or so we will go with option 2, crediting you — but we would rather it stayed your PR if you are still interested.

insolace:
Fixed in 389, thank you @franko.

PR #347, opened 2025-02-20
Android: prevent missing PROPERTY_NAME from causing a nullptr dereference.

conradhappeliv:
Not all MidiDeviceInfo structs have the PROPERTY_NAME property. To work around this, this will fall back to PROPERTY_PRODUCT, and just return an empty string if that's also missing.


insolace:
I'm going to build/test this with hardware to verify before merging

insolace:
Reproduced and confirmed this is the correct fix. Thank you @conradhappeliv

PR #350, opened 2025-04-15
warning: variable length arrays in C++ are a Clang extension

JlnWntr:
warning: variable length arrays in C++ are a Clang extension) extension]  1686 |   Byte buffer[bufsize+16]; // pad for other struct members

I compiled this with `g++` (clang, std c++17) on a recent macOS command line.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -1683,8 +1683,9 @@ void MidiOutCore :: sendMessage( const unsigned char *message, size_t size )
   OSStatus result;
 
   ByteCount bufsize = nBytes > 65535 ? 65535 : nBytes;
-  Byte buffer[bufsize+16]; // pad for other struct members
-  ByteCount listSize = sizeof( buffer );
+  ByteCount listSize = (bufsize+16) * sizeof(Byte);
+  Byte *buffer = (Byte *) alloca(listSize);
+  
   MIDIPacketList *packetList = (MIDIPacketList*)buffer;
 
   ByteCount remainingBytes = nBytes;


keinstein:
I think this workaround should use `vector` instead of C-style arrays. This would use would automatically cleanup the allocated memory. As an alternative alloca and related functions could be used to put this array on the stack, so that it gets cleaned when the function is left in any way.

jonathon-bell:
First, thank you all for maintaining such a useful project.

I too see this warning using clang-20 on macOS. It's annoying, though harmless.

- @keinstein I respectfully disagree that switching to an `std::vector` is the best fix here: its certainly would address the problem, but in my view is overkill because it introduces an essential use of the heap, which could possibly affect performance for what may be a performance critical section of code.

- that leaves the stack: note that the function does not recurse, so the stack space consumed by placing the buffer on the stack is bounded and in the order of at most ~64k.

- alloca also addresses the issue, but is a posix API, not standard C++. However this section of the code is already tied to the macOS API, so perhaps is the best fix.

- another possibility, however, would be to simply ignore the `size` parameter and fix the buffer size at the maximum possible size of `64k+16`. Yes, for many messages this is more than is needed, but....does that really matter? It seems a relatively small amount of stack memory to consume in the overall scheme of things, (unless perhaps compiling for an embedded system; does RtMidi support such platforms?)

- yet another possibility is to simply disable the warning with a #pragma and leave the compiler to do it's thing.

Please let me know if I can be of help in putting together a PR - would be happy to do so.

Regards, Jonathon

insolace:
Picking this up now that I have dev/commit access. Thank you both for the analysis!

@jonathon-bell, your objection to std::vector is fair and well reasoned, but given some of the other issues that have come up related to sendMessage, I think vector is the better solution. The allocation here happens once per sendShort() invocation, which is control-rate traffic - a SysEx dump or an identity request - not a per-sample or per-note path. A single heap allocation of up to ~64 KB is small next to the CoreMIDI and USB transfer latency that follows it. So the performance concern is real in principle but, when testing against our USB products, in practice isn't a significant factor at this call site. If someone measures otherwise on real hardware I would genuinely want to know.

We've done a ton of work and testing to get our large sysex/firmware update payloads to work consistently across each backend, and for us the deciding factor is failure mode rather than speed. alloca() of ~64 KB happens on a thread RtMidi does not own - the caller's thread, with a stack budget RtMidi cannot see. If that budget is short, alloca does not fail in any catchable way; it corrupts or crashes. std::vector throws std::bad_alloc, which an application can handle. For a library whose whole job is to be dropped into someone else's process, I would rather fail loudly than unpredictably.

On your other two suggestions: a fixed 64 KB + 16 stack buffer has the same unowned-stack problem, and RtMidi does target constrained platforms (we use it on a few of our embedded products), so I would rather not assume the headroom. Suppressing with #pragma leaves a real portability issue in place — VLAs are a compiler extension, and MSVC doesn't support them, so the code is only warning-free by accident of which compiler you use.

@JlnWntr, I opened #381 with the std::vector version rather than updating yours, which I should have offered first, my mistake. If you would rather update your PR I'd be happy to merge it so you get the credit, say so and I will close #381 in its favour.

@jonathon-bell - your offer to help is much appreciated. I've just come on board as a developer and have started to organize issues/PRs, and have created two milestones for a pending 6.1.0 release to fix pressing issues, and a much more involved 7.0.0 release to add MIDI 2.0 support. I'll be adding some tracking issues to help plan these efforts, there is a good deal of work to be done if you're still interested.

JlnWntr:
@insolace, it's done. Thanks for the credit.

PR #357, opened 2025-08-17
Add support for C++20 modules

mikomikotaishi:
This pull request adds support for C++20 modules, as well as puts all symbols in the inlined namespace `rtmidi`. This change does not break any existing code, but allows symbols to be optionally accessed through the namespace `rtmidi` in addition to the global namespace. The module itself is imported using `import rtmidi;` and is enabled with the build macro `RTMIDI_BUILD_MODULES`.


mikomikotaishi:
I think it might be worth looking in to having the inline namespace be optional, so with a compiler flag it could be a mandatory namespace for those who want the symbols strictly outside of the global namespace

mikomikotaishi:
May I please get a response or some kind of feedback

PR #359, opened 2025-08-23
Add `rtmidi_set_error_callback` to C wrapper

adsr:
Without setting a callback, all errors are printed to stderr, which may not be desirable.

Rename `data` field in wrapper struct to something more descriptive.

Delete `rtmidi_error` function as the prototype mismatched the implementation.

Templatize existing callback proxy class so we can reuse it for error callbacks.

This is an ABI break as it adds a field to the wrapper struct.

Test script:

```c
#include <stdio.h>
#include <rtmidi_c.h>

static void error_cb(enum RtMidiErrorType type, const char *msg, void *udata) {
    printf("error_cb: type=%d msg=%s udata=%p\n", type, msg, udata);
}

int main(int argc, char **argv) {
    RtMidiOutPtr out = rtmidi_out_create(0, argv[0]);
    rtmidi_set_error_callback(out, error_cb, (void *)0xffff);
    rtmidi_open_port(out, 999, "invalid_port");
    rtmidi_out_free(out);
    return 0;
}
```

```console
$ ./test
error_cb: type=6 msg=MidiOutAlsa::openPort: the 'portNumber' argument (999) is invalid. udata=0xffff
```

### DIFF
--- a/rtmidi_c.cpp
+++ b/rtmidi_c.cpp
@@ -34,14 +34,15 @@ class StaticEnumAssertions { StaticEnumAssertions() {
     ENUM_EQUAL( RTMIDI_ERROR_THREAD_ERROR,       RtMidiError::THREAD_ERROR );
 }};
 
+template <typename T>
 class CallbackProxyUserData
 {
   public:
-  CallbackProxyUserData (RtMidiCCallback cCallback, void *userData)
+  CallbackProxyUserData (T cCallback, void *userData)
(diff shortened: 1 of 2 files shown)


garyscavone:
There is a conflict, probably with the previous PR merge. Please see if you can resolve it, as I don't use the C API at all and it would take a while for me to evaluate the conflict.

PR #361, opened 2025-09-25
Fix build error caused by use of deprecated POSIX name of `strdup`

Nytra:
Changes `strdup` to `_strdup`

Fixes the following build error on Windows 11 with VS2022:

```
rtmidi_c.cpp(420,19): error C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _strdup. See online help for details.
```

### DIFF
--- a/rtmidi_c.cpp
+++ b/rtmidi_c.cpp
@@ -417,5 +417,5 @@ static void rtmidi_set_error_msg (RtMidiPtr device, const char *err)
     if (device->msg) {
         free (device->msg);
     }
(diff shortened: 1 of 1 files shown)


keinstein:
Ironically C24 has adopted the strdup functiion as mentioned in the [https://blog.ansi.org/ansi/c-programming-language-standard-iso-iec-9899-2024/](ANSI C Blog) so _strdup will probably be deprecated in the far future.

PR #372, opened 2026-05-11
WinMM: avoid deadlock

jkeller51:
I was constantly experiencing a deadlock in my application. It seems when calling `midiInReset()`, Windows will try to finish processing any remaining input MIDI through `midiInputCallback()`. In `closePort()`, the mutex is acquired, then `midiInReset()` is called. If there is any remaining sysex to process, `midiInputCallback()` will attempt to acquire the mutex from the windows audio thread, leading to a deadlock.

This PR attempts to avert the deadlock by setting a closing flag from `closePort()`, which will bypass the sysex processing (and mutex acquisition) in  `midiInputCallback()`.


insolace:
Heads up — we ran into this same deadlock while fixing #376 and came up with the same approach, before spotting your PR. Sorry for the duplicate effort.

Our version is in #380, but it is bundled with a larger input teardown rewrite, so yours is the smaller and more obviously reviewable fix for the deadlock on its own. Happy to rebase ours on top if this one lands first.

For what it is worth, we reproduced the crash on real hardware here: `MidiInWinMM::closePort` killed the process with `0xC0000409` four times in a day, mostly after a device rebooted mid-transfer.

insolace:
@jkeller51 I'd like to merge this so you get credit and then rebase #380 on top of it, but your PR is failing the MinGW CI check. Would you mind updating it? 

**The CI failure is a one-line fix.** Both MinGW jobs fail while Linux, JACK and macOS pass, and the cause is that `std::atomic<bool> closing;` needs `#include <atomic>`. RtMidi does include it, but at line 368, inside the `#if defined(__AMIDI__)` block, so it is only visible on Android builds. MSVC and the Linux/macOS toolchains pull `<atomic>` in transitively through other headers; MinGW's libstdc++ does not, which is why it fails only there.

Adding the include inside the WinMM block, near the other Windows includes, should be enough. Pushing any commit to this branch will also re-trigger the checks, which is the simplest way to get a fresh run.

jkeller51:
Okay, I've fixed the issue.

Issue #376, opened 2026-09-08
MidiInWinMM::closePort leaks the critical section, leaves connected_ true and double-frees the sysex buffers when midiInUnprepareHeader fails


Psychlist1972:
Hi

I had Copilot/Claude do a survey of WinMM / WinRT MIDI 1.0 issues and practices, based on common bugs we find in apps and libraries, and failure modes reported to us. I'm filing bugs with the findings. The intent here is to help library authors make MIDI better for everyone on Windows.

I know this repo hasn't seen much action lately, but documenting issues here in the hopes that any forks or other downstream libraries/users will know about these issues.

Everything after the break was reported by GitHub Copilot (Claude Opus 5)

Pete
Microsoft

---

**Repo:** thestk/rtmidi
**Version:** `master` @ `a3233c22949342f6697681e2cf2403e27fcf0c9e` (2025-11-27). Also present in 6.0.0.
**File:** `RtMidi.cpp`, `MidiInWinMM`
**Backend:** `__WINDOWS_MM__`, which is the only Windows backend the build system can select

---

## Summary

`MidiInWinMM::closePort()` has an error path for `midiInUnprepareHeader` failing. That path frees the buffers before it looks at the result, then returns from the middle of the function without releasing the critical section it is holding and without clearing `connected_`.

Because `connected_` is still true, the destructor calls `closePort()` a second time, and the second pass walks the same `sysexBuffer` vector, which still holds the pointers that were already deleted.

I am not reporting an observed crash, and I have not established which conditions make `midiInUnprepareHeader` fail here. I do not think I need to: the function already assumes it can fail, because it checks the result and writes an error path for it. This report is about that error path being unsafe rather than about how often it runs.

## The code

[`RtMidi.cpp#L2916-L2942`](https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/RtMidi.cpp#L2916-L2942):

```cpp
void MidiInWinMM :: closePort( void )
{
  if ( connected_ ) {
    WinMidiData *data = static_cast<WinMidiData *> (apiData_);
    EnterCriticalSection( &(data->_mutex) );
    midiInReset( data->inHandle );
    midiInStop( data->inHandle );

    for ( size_t i=0; i < data->sysexBuffer.size(); ++i ) {
      int result = midiInUnprepareHeader(data->inHandle, data->sysexBuffer[i], sizeof(MIDIHDR));
[...]

PR #378, opened 2026-09-10
Small fixes: uninstall target guard, signed/unsigned compare, dead link

### LINKED ISSUE #291: Warning Implicit conversion loses integer precision:
I'm now using macOs / XCode. it is not an issue but a suggestion of supress some c++ warnings.
```
Implicit conversion loses integer precision: 'ItemCount' (aka 'unsigned long') to 'unsigned int'
```
https://github.com/thestk/rtmidi/blob/454e1ce473ef3436fae70eca23623d267d4fbac7/RtMidi.cpp#L1066
https://github.com/thestk/rtmidi/blob/454e1ce473ef3436fae70eca23623d267d4fbac7/RtMidi.cpp#L1177
https://github.com/thestk/rtmidi/blob/454e1ce473ef3436fae70eca23623d267d4fbac7/RtMidi.cpp#L1273
https://github.com/thestk/rtmidi/blob/454e1ce473ef3436fae70eca23623d267d4fbac7/RtMidi.cpp#L1397
https://github.com/thestk/rtmidi/blob/454e1ce473ef3436fae70eca23623d267d4fbac7/RtMidi.cpp#L1433

I can make a PR if needed but preferred to open an issue to see how it is best.
maybe cast to data types like this?
```c++
  unsigned int nDest = (unsigned int)MIDIGetNumberOfDestinations();
```
cheers

### LINKED ISSUE #369: RtMidi is creating target "uninstall" without checking if such a target exists and causing CMake config to fail
As stated in the title, RTMidi is creating a target `uninstall` without checking if such a target exists and causing CMake config to fail.

The other 3rd party libs I'm linking against that also try to create the target `uninstall` check first if such a target already exists, which might be what needs to happen in the RTMidi CMake too. This wasn't a problem until I upgraded CMake to 4.1 so I'm ready for the impending C++26... Apparently before that it was just silently ignoring the conflict or something.

Incase anyone else encounters this, my messy workaround currently is using this patch command to string replace a guard into your CMake during CPM package installation:

I have the following patch script `${PROJECT_SOURCE_DIR}/cmake/patch_rtmidi.cmake`:

```
file(READ "${RTMIDI_SOURCE_DIR}/CMakeLists.txt" content)

string(REPLACE
        "add_custom_target(\${RTMIDI_TARGETNAME_UNINSTALL}"
        "if(NOT TARGET \${RTMIDI_TARGETNAME_UNINSTALL})\nadd_custom_target(\${RTMIDI_TARGETNAME_UNINSTALL}"
        content "${content}")

string(REPLACE
        "COMMAND \${CMAKE_COMMAND} -P \${CMAKE_BINARY_DIR}/RtMidiConfigUninstall.cmake)"
        "COMMAND \${CMAKE_COMMAND} -P \${CMAKE_BINARY_DIR}/RtMidiConfigUninstall.cmake)\nendif()"
        content "${content}")

file(WRITE "${RTMIDI_SOURCE_DIR}/CMakeLists.txt" "${content}")
```

And the modified CPM command to patch after it downloads the repo is as follows:

```
CPMAddPackage(
        NAME rtmidi
        GIT_REPOSITORY https://github.com/thestk/rtmidi
        GIT_TAG 5.0.0
        PATCH_COMMAND ${CMAKE_COMMAND} -DRTMIDI_SOURCE_DIR=<SOURCE_DIR> -P ${PROJECT_SOURCE_DIR}/cmake/patch_rtmidi.cmake
)
```

### LINKED ISSUE #371: Dead link to Google Groups in WinMM implementation
There is a dead link to a Google Groups discussion in the WinMM implementation of RtMidi:

https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/RtMidi.cpp#L2690

It's not possible to retrieve this URL via archive.org (it just normalizes the URL to `https://groups.google.com/forum/`, due to the actually important bits of the URL being placed in an anchor). I'd like to know what this is about and why the Mutex is necessary in some of the places it's used. Does anyone remember?

insolace:
Three small unrelated fixes, each closing an open issue.

**#369** — `add_custom_target(uninstall)` is created unconditionally, which fails configuration when RtMidi is added as a subproject alongside another library that defines its own `uninstall` target. Previously tolerated; fatal as of CMake 4.1. Guarded with `if(NOT TARGET ...)`.

**#291** — `CreateConnectedEndpointName` declares `int i` and compares it against `size_t nConnected`. The `nConnected` half was already changed to `size_t` upstream; this changes `i` to match.

**#371** — the mididev Google Groups thread referenced in `WinMidiData` is no longer publicly reachable (the group returns an access error). Replaced the link with what it explained, keeping the attribution.

Verified on Linux (ALSA + JACK), CMake configure and build clean. The `#291` change is in CoreMIDI code, which does not compile on Linux — worth a look from someone on macOS.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -1399,7 +1399,7 @@ static CFStringRef CreateConnectedEndpointName( MIDIEndpointRef endpoint )
   CFMutableStringRef result = CFStringCreateMutable( NULL, 0 );
   CFStringRef str;
   OSStatus err;
-  int i;
+  size_t i;
 
   // Does the endpoint have connections?
   CFDataRef connections = NULL;
@@ -2687,7 +2687,7 @@ struct WinMidiData {
   DWORD lastTime;
   MidiInApi::MidiMessage message;
   std::vector<LPMIDIHDR> sysexBuffer;
-  CRITICAL_SECTION _mutex; // [Patrice] see https://groups.google.com/forum/#!topic/mididev/6OUjHutMpEo
+  CRITICAL_SECTION _mutex; // [Patrice] protects the sysex buffer requeue in midiInputCallback
 };
 
 //*********************************************************************//
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,8 +294,13 @@ configure_file(
     "${CMAKE_BINARY_DIR}/RtMidiConfigUninstall.cmake" @ONLY)
 
 # Create uninstall target.
-add_custom_target(${RTMIDI_TARGETNAME_UNINSTALL}
-    COMMAND ${CMAKE_COMMAND} -P ${CMAKE_BINARY_DIR}/RtMidiConfigUninstall.cmake)
+# Guard against the target already existing: when RtMidi is added as a
+# subproject alongside another library that defines its own 'uninstall'
+# target, redefining it is an error (fatal as of CMake 4.1).
+if(NOT TARGET ${RTMIDI_TARGETNAME_UNINSTALL})
+  add_custom_target(${RTMIDI_TARGETNAME_UNINSTALL}
+      COMMAND ${CMAKE_COMMAND} -P ${CMAKE_BINARY_DIR}/RtMidiConfigUninstall.cmake)
+endif()
 
 install(
     FILES ${CMAKE_CURRENT_BINARY_DIR}/rtmidi.pc


insolace:
Follow-up on verification: the `int i` -> `size_t i` change in `CreateConnectedEndpointName()` is CoreMIDI code, so it had not been compiled when this was opened. It has now been built and exercised on macOS (Clang, no warnings) against real MIDI hardware — MIDI-CI Discovery and several Property Exchange transactions, all of which go through that port-name comparison — with results byte-identical to the unpatched baseline.

The other two changes (the CMake `uninstall` target guard and the comment edit) were verified on Linux when this was opened.

RtMidi.cpp
```diff
@@ -2687,7 +2687,7 @@ struct WinMidiData {
   DWORD lastTime;
   MidiInApi::MidiMessage message;
   std::vector<LPMIDIHDR> sysexBuffer;
-  CRITICAL_SECTION _mutex; // [Patrice] see https://groups.google.com/forum/#!topic/mididev/6OUjHutMpEo
+  CRITICAL_SECTION _mutex; // [Patrice] protects the sysex buffer requeue in midiInputCallback
```
sagamusix, line 2690: I'm not really sure a comment like this is needed. It explains what the mutex is doing, which is obvious from just looking at where the mutex is locked and unlocked. What I was hoping for to find in the dead Google Groups link is the *why*: How can it be possible that this buffer is accessed from multiple threads, necessitating a mutex? This question remains unanswered.

insolace: The buffer is reachable from two threads because midiInOpen() is called with CALLBACK_FUNCTION, so WinMM invokes midiInputCallback() on a thread it owns, not on any thread the application controls. That callback requeues the sysex buffer with midiInAddBuffer(). Meanwhile MidiInWinMM::closePort() runs on the application's thread and walks the same sysexBuffer vector, calling midiInUnprepareHeader() and freeing each header. Those are the only two places the critical section is taken, and they are the two sides of the race: the driver thread requeueing a buffer that the application thread is in the middle of retiring.

midiInReset() makes it concrete. closePort() calls it while holding the lock, and it is precisely what completes pending buffers and drives the callbacks that then want the same lock. So the mutex is not protecting against two application threads, it is protecting against the driver's own callback thread arriving during teardown.

I could update the comment with something saying that midiInputCallback runs on a WinMM-owned thread and requeues these buffers while closePort retires them from the caller's thread, and this guards that overlap.

Also the surrounding code has active work: #372 fixes a deadlock on exactly this interaction, and #380 rewrites the input teardown for #376.

insolace:
Updated the comment, onward and upward.

Issue #214, opened 2020-03-31
ALSA and JACK do not support large SysEx messages


keinstein:
As there were some patches that improve the support of larger SysEx messages on different backends  (e.g. #91 ). However neither ALSA nor JACK support these kind of messages. While the ALSA sequencer theoretically allows larger SysEx messages. The kernel silently clears the input buffer as soon as an event could not be delivered to it. This means when the sender is to fast by a single packet about 2-3k of SysEx data gets silently dropped, when communicating with a software device.

A workaround could be to implement RawMIDI support for ALSA.
Another workaround could be to implement user defined input/output pool sizes.

KoSv:
hi there, 
I'm trying to send around 6k and my hardware device freezes (sysex massage is not complete)
where receiving is not a problem!
Is there a workaround?
greets and happy christmas! :)

jcelerier:
Hello,
I have a partial work on a raw midi backend with support for chunking on the https://github.com/jcelerier/libremidi fork, testing would be appreciated.

insolace:
Hello,

I've confirmed that the libremidi fork works. The RAW send method can send large (60-100k) sysex payloads (firmware), the current RtMidi ALSA send method cannot handle more than 1-2k messages.

jcelerier:
yay :)

insolace:
> yay :) 

@jcelerier Any chance you'd be willing to port your fix into rtmidi and make a pull request?

jcelerier:
Well, it's more of a complete rewrite than a fix, I just don't have enough time for that sorry :(

insolace:
I figured as much when I looked at your Raw Send code. I wish I had the time to do this myself, but for now we will have to hold off on releasing Linux versions of our editors until this issue is fixed.

keinstein:
@jcelerier Did I read libremidi right that you have the possibility to configure the transfer rate by inserting an artificial sleep in the ALSA sending routine? This could be a compromise which is much easier to backport.

jcelerier:
yes pretty much, but that's with the "raw" alsa api, not the seq one (I remember trying and failing to make it work there though I could just have missed something simple... that was some months ago, it's not entirely clear in my mind). All the relevant code is there, feel free to backport: https://github.com/jcelerier/libremidi/blob/master/include/libremidi/detail/raw_alsa.hpp#L447

PR #379, opened 2026-09-11
ALSA: accept partial SysEx spans in MidiOutAlsa::sendMessage

insolace:
ALSA is the only backend that refuses a SysEx fragment.

`MidiOutAlsa::sendMessage()` re-encodes its bytes through `snd_midi_event_encode()`, which only emits an event for a self-contained message. Hand it a fragment of a larger SysEx and it returns `SND_SEQ_EVENT_NONE`, which the send loop treats as fatal — the bytes are dropped and a WARNING is raised that most callers never see.

CoreMIDI, WinMM, JACK and Web all forward SysEx bytes to the transport as given, so a caller may hand a large SysEx to `sendMessage()` across several calls — an opening `F0..` with no `F7`, bare continuation bytes, then a span ending in `F7` — and have it arrive. Pacing a transfer that way is an established idiom for firmware updates, where the inter-span gaps need to land as real gaps on the wire. On ALSA the same code silently loses everything up to the final span.

## Reproduction

Against alsa-lib 1.2.14 on a real MIDI port, comparing this branch to master at a3233c2. An opening span with no `F7`:

| span | master | this branch |
|---|---|---|
| 4096 B | `incomplete message!` | ok |
| 8192 B | `incomplete message!` | ok |
| 10240 B | `incomplete message!` | ok |

## The change

Track SysEx state on `AlsaMidiData` and, when a call opens or continues a SysEx, emit the bytes with `snd_seq_ev_set_sysex()` instead of running them through the encoder. That is ALSA's native transport for a large SysEx — a sequence of SysEx events carrying arbitrary byte spans, concatenated by the receiver until `F7` — and the mirror of the reassembly `MidiInAlsa` already performs.

Any send failure clears the framing state, and so does `closePort()`: a caller that errors part-way through abandons that message, and a stale in-progress flag would route its next, unrelated, message down this path instead of the encoder.

Anything that is not part of a SysEx takes the original encoder path, unmodified. The diff is 55 insertions and 0 deletions; existing behaviour for channel, system-common and realtime messages is untouched.

No public API change — `RtMidi.h` is not touched.

## What this does not fix

A **single** message larger than roughly 16 KB is refused by `snd_seq_event_output()` with `-EINVAL`, on master and with this change alike. That is a sequencer output-pool limit rather than an encoder one — the exact ceiling tracks the client's output pool size (16355 bytes with the default 500-event pool on the test machine), so it is closer to the input-side concern originally reported in #214.

Streaming a large SysEx as sub-ceiling spans, which this change enables, is a way around it. Raising the ceiling itself — `snd_seq_set_client_pool_output()`, or having `sendMessage()` split internally — is separate work, and I have deliberately not done it here. Happy to follow up if you have a preference on direction.

## Scope

Addresses the ALSA **output** path of #214. The >16 KB single-message ceiling, the input-side kernel buffering keinstein originally described, and the JACK half all remain open.

Adjacent, no conflict expected: #374 touches the same file in the input reconnection path, not `MidiOutAlsa::sendMessage`.

Tested on Linux (ALSA + JACK): Autotools (`autogen.sh`, `configure --with-alsa`, `make`, `make check`) and CMake both configure and build clean, and a 566-chunk / 63 KB firmware update to real hardware completes with no ALSA diagnostics.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -2598,15 +2598,20 @@ void MidiOutAlsa :: sendMessage( const unsigned char *message, size_t size )
   if ( nBytes > 0 ) {
     const unsigned char first = message[0];
     if ( data->sysexInProgress || first == 0xF0 ) {
-      // Track completion so the next call knows whether it is a continuation.
+      // Work out the framing state this buffer leaves behind, but commit it
+      // only once the event has been accepted (below). If the send fails, the
+      // state stays as it was, so a caller that retries the same span gets it
+      // routed the same way; otherwise a retried continuation would fall
+      // through to the encoder and be rejected as an incomplete message.
       // Real-time bytes (0xF8-0xFF) may appear mid-SysEx without ending it;
       // F7 ends it, and so does any other non-real-time status byte.
+      bool nextSysexInProgress = data->sysexInProgress;
       for ( unsigned int i = 0; i < nBytes; ++i ) {
         const unsigned char b = message[i];
         if ( b >= 0xF8 ) continue;
-        if ( b == 0xF0 ) { data->sysexInProgress = true; continue; }
-        if ( b == 0xF7 ) { data->sysexInProgress = false; continue; }
-        if ( b >= 0x80 ) data->sysexInProgress = false;
+        if ( b == 0xF0 ) { nextSysexInProgress = true; continue; }
+        if ( b == 0xF7 ) { nextSysexInProgress = false; continue; }
+        if ( b >= 0x80 ) nextSysexInProgress = false;
       }
 
       snd_seq_event_t ev;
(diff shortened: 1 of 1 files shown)


insolace:
Pushed a follow-up commit (a9c07c6): any send failure now clears the SysEx framing state, and so does `closePort()`.

Without it, a caller that errors part-way through a fragmented SysEx left `sysexInProgress` set, so its next unrelated message was routed down the SysEx path instead of through the encoder. Same fix is being applied to the equivalent WinMM framing state in separate work, so the two backends stay consistent.

No change to the behaviour described above; still 0 deletions.

insolace:
Before I merge my unreviewed code, would anyone with ALSA / SysEx experience be willing to take a look at this and provide some comments?

@madbrain76
@keinstein
@SpotlightKid

madbrain76:
I don't have any objection.

Issue #375, opened 2026-09-08
UWP backend: init_apartment() runs in a static initializer, and the port list is a snapshot that never refreshes


Psychlist1972:
Hi

I had Copilot/Claude do a survey of WinMM / WinRT MIDI 1.0 issues and practices and I'm filing bugs with the findings. The intent here is to make MIDI better for everyone on Windows.

I know this repo hasn't seen much action lately, but documenting issues here in the hopes that any forks or other downstream libraries/users will know about these issues.

Everything after the break was reported by GitHub Copilot (Claude Opus 5)

Pete
Microsoft

---

**Repo:** thestk/rtmidi
**Version:** `master` @ `a3233c22949342f6697681e2cf2403e27fcf0c9e` (2025-11-27)
**File:** `RtMidi.cpp`, the `__WINDOWS_UWP__` section
**Priority:** low, and I would like to explain why up front

---

## Reachability first, because it changes how much any of this matters

`__WINDOWS_UWP__` is not selectable from the build system. `CMakeLists.txt` defines `RTMIDI_API_WINMM` and emits `-D__WINDOWS_MM__` at [`CMakeLists.txt#L139-L143`](https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/CMakeLists.txt#L139-L143), and `configure.ac` does the equivalent, but neither has any option that defines `__WINDOWS_UWP__`. The only way to compile this backend is to define the macro by hand.

So nothing downstream is running this code, and none of what follows is hurting anyone today. I am reporting it because untested code that is one `-D` away from being live is worth knowing about, and because the first item would be difficult to diagnose if someone did turn it on.

## `init_apartment()` runs during static initialization

[`L3249-L3257`](https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/RtMidi.cpp#L3249-L3257) and [`L3396`](https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/RtMidi.cpp#L3396):

```cpp
class UWPMidiInit
{
public:
    UWPMidiInit()
    {
        winrt::init_apartment();
    }
};
...
UWPMidiInit UWPMidiClass::uwp_midi_init_;     // namespace-scope definition
```

[...]

Issue #333, opened 2024-03-06
midiOutUnprepareHeader intermittently hangs on Windows during repeated large sysex sends


insolace:
I tried submitting a pull request but did not have permissions. 

Starting at line 3184 of RtMidi.cpp:

// Unprepare the buffer and MIDIHDR.
    while ( MIDIERR_STILLPLAYING == midiOutUnprepareHeader( data->outHandle, &sysex, sizeof ( MIDIHDR ) ) ) Sleep( 1 );

This line of code is problematic and can cause our application to crash during our firmware update process, when we are sending multiple reboot commands in between large sysex payloads. Our current workaround is to delete the RtMidi in/out objects and then recreate them, otherwise depending on the state of the update process things can hang.

We'd prefer if there was better error handling and/or a timeout inside of this loop. Here is a simple timeout that would be better if there was a method to change the duration:

    int timeOut = 0;
    // Unprepare the buffer and MIDIHDR.
    while ( MIDIERR_STILLPLAYING == midiOutUnprepareHeader( data->outHandle, &sysex, sizeof ( MIDIHDR ) ) )
    {
        if (timeOut++ > 25000)
            break;
        Sleep( 1 );
    }

insolace:
It looks like another issue references this loop but did not encounter the same bug as I am:

https://github.com/thestk/rtmidi/issues/250

Debugging this further, my suggested timeout code does not actually prevent the application from hanging/crashing. It would seem that having a device reboot after receiving a large sysex payload can create situations where calling midiOutUnprepareHeader hangs, and there's no method in the API to check the validity of the handle or the buffer before calling it. 

Our current workaround is to send the firmware, wait for the midi ports to change to confirm that the device has rebooted, and then stop the application with a user dialog/ok that then kills and relaunches the application. Otherwise if we try to send sysex at that point, there's a 50/50 chance that calling midiOutUnprepareHeader will crash the application.

Issue #234, opened 2021-02-09
Cannot open certain port on Windows


crankedguy:
Hello, I opened this issue on python-rtmidi as I have problems opening a certain Midi port with rtmidi which works flawlessly in other applications.

https://github.com/SpotlightKid/python-rtmidi/issues/78

As the issue happens in the rtmidi code here 
https://github.com/thestk/rtmidi/blob/84d130bf22d878ff1b0e224346e2e0f9e3ba8099/RtMidi.cpp#L2604
I decided to ask also here about the issue

Basically you have the following ports 
D:\Development\Push2\push2-captureone>python try_rtmidi.py
Bome MIDI Translator 1 0
Ableton Push 2 1
MIDIIN2 (Ableton Push 2) 2

You can open the first 2 ports without any issue but if you want to open port 2 "MIDIIN2..." you get the following error from rtmidi
Traceback (most recent call last):
File "D:\Development\Push2\push2-captureone\try_rtmidi.py", line 10, in
midiin.open_port(2)
File "src/_rtmidi.pyx", line 586, in rtmidi._rtmidi.MidiBase.open_port
File "src/_rtmidi.pyx", line 220, in rtmidi._rtmidi._cb_error_func
File "src/_rtmidi.pyx", line 408, in rtmidi._rtmidi._default_error_handler
rtmidi._rtmidi.SystemError: MidiInWinMM::openPort: error creating Windows MM MIDI input port.

Hope you can help with that as I cannot see any issue with this port. It works, sends and receives data.

Thanks

SpotlightKid:
My suspicion would be that it has to do with the parentheses in the port name, but I'm only guessing.

crankedguy:
That was actually my first guess when I saw it, that this goes into the direction of a parsing error/not allowed chars in the name

SpotlightKid:
It would maybe help, if RtMidi included the [the error code in the return value](https://docs.microsoft.com/en-us/windows/win32/api/mmeapi/nf-mmeapi-midiinopen#return-value) of WinMM midiInOpen in its error message.

jean-emmanuel:
I'm also interested in this, having some users on Windows reporting similar issues (although not necessarely related to parentheses in port names). Any chance this is related to the limitation mentionned in https://github.com/thestk/rtmidi/issues/102 ?

crankedguy:
[...]

Issue #102, opened 2017-02-17
RtMidi prevents another application to read/write to opened device


sinosoidal:
Hi,

I'm having an issue with RtMidi that I don't know if it is a limitation or something I'm failing to prevent. I have an app that lists all available midi devices. During this process I open each one of them with RtMidi. When this app is running and if I open another audio application, let's say, Ableton Live, the devices become unavailable for Ableton Live. When RtMidi opens a device, does it block the access by other audio/midi applications?

Thx

radarsat1:
I don't see anything in particular in MidiInCore :: openPort() (RtMidi.cpp) that is setting any exclusivity flags, but I'm unfamiliar with the expected behaviour.  Are you opening them and keeping them open?  Or do you mean that they stay reserved after you close them? Is it normal for more than one application to be able to open the same CoreMIDI port?  Is it for reading or writing?  (receiving midi or sending midi?)

For reference, the following functions are called during openPort()
* MIDIInputPortCreate
*  MIDIGetSource( portNumber )
* MIDIPortConnectSource

sinosoidal:
The problem is that if I instantiate a RtMidi In or Out and open the port, it becomes unavailable on Ableton Live. Curiously only the external usb midi devices ports become unavailable, since the virtual (loopMidi) and the Microsoft GS Wavetable Synth are kept available.

Yes, it is supposed for several different midi applications to have a access to midi ports simultaneously.

This is what I do in order to open a specific in port in this case:

    void IMidiDevice::openIn()
    {
        RtMidiIn *port = new RtMidiIn();

        uint i=0;

        for (; i<port->getPortCount(); i++)
        {
            if (QString::fromUtf8(port->getPortName(i).c_str()).compare(_name)==0)
        }

        if (i<port->getPortCount())
        {
            try
            {
                  port->openPort(i);
                  port->setCallback(&midiInCallback, (void*)this);
                  port->ignoreTypes(false, false, false);
                  _in = port;
            }
            catch (RtMidiError &error)
            {
                Q_UNUSED(error)
            }
        }
    }

radarsat1:
I can't answer to,

> Yes, it is supposed for several different midi applications to have a access to midi ports simultaneously.

as I can't see this in any documentation, but perhaps it's the expected OS X user experience, I don't know.  In that case you'd need to find out if there is an exclusivity flag to set or unset somewhere in the CoreMIDI API.

[...]

PR #380, opened 2026-09-11
WinMM: fix SysEx short-tail routing, output buffer lifetime, and input closePort teardown

### LINKED ISSUE #325: MidiInWinMM :: closePort might return without calling LeaveCriticalSection
During a code review I noticed that the method MidiInWinMM :: closePort looks like it might possibly fail to release an acquired critical section:

```
void MidiInWinMM :: closePort( void )
{
  if ( connected_ ) {
    WinMidiData *data = static_cast<WinMidiData *> (apiData_);
    EnterCriticalSection( &(data->_mutex) );
    midiInReset( data->inHandle );
    midiInStop( data->inHandle );

    for ( size_t i=0; i < data->sysexBuffer.size(); ++i ) {
      int result = midiInUnprepareHeader(data->inHandle, data->sysexBuffer[i], sizeof(MIDIHDR));
      delete [] data->sysexBuffer[i]->lpData;
      delete [] data->sysexBuffer[i];
      if ( result != MMSYSERR_NOERROR ) {
        midiInClose( data->inHandle );
        data->inHandle = 0;
        errorString_ = "MidiInWinMM::openPort: error closing Windows MM MIDI input port (midiInUnprepareHeader).";
        error( RtMidiError::DRIVER_ERROR, errorString_ );
        return;
      }
    }

    midiInClose( data->inHandle );
    data->inHandle = 0;
    connected_ = false;
    LeaveCriticalSection( &(data->_mutex) );
  }
}
```

At the end of the for loop, return might be called, and this would exit the function without calling LeaveCriticalSection on the last line.
Is this intentional, or a bug? If it is a bug, using a Locker class that will call LeaveCriticalSection in its destructor would fix the issue.

### LINKED ISSUE #373: WinMM: MidiOutWinMM::sendMessage makes potentially incorrect assumptions based on data size
Some companies use this code to send SysEx messages for things like firmware updates. If the last chunk of SysEx they send over happens to be 3 or fewer bytes, then the code for WinMM rtMIDI takes an incorrect branch and sends the SysEx through `midiOutShortMsg`, which shall not be used for that purpose.

The end result is the remaining 1-3 bytes of SysEx are not actually sent to the device.

Relevant code

```cpp
 if ( nBytes > 3 ) { // Sysex message
 }
  else { // Channel or system message.
 }
```

There are a number of ways this could be fixed, depending on what's normal for this library. Here are two ideas off the top of my head:
- Add a "continued sysex" or other flag as the last parameter of the function in an overload
- If 3 bytes or fewer, check to see if the first byte is a data byte or sysex end status, and if so, send via `midiOutLongMsg` instead

Pete
Microsoft

### LINKED ISSUE #376: MidiInWinMM::closePort leaks the critical section, leaves connected_ true and double-frees the sysex buffers when midiInUnprepareHeader fails

### LINKED ISSUE #377: MidiOutWinMM::sendMessage frees a prepared buffer without unpreparing it whenever midiOutLongMsg or midiOutUnprepareHeader fails
Hi

I had Copilot/Claude do a survey of WinMM / WinRT MIDI 1.0 issues and practices, based on common bugs we find in apps and libraries, and failure modes reported to us. I'm filing bugs with the findings. The intent here is to help library authors make MIDI better for everyone on Windows.

I know this repo hasn't seen much action lately, but documenting issues here in the hopes that any forks or other downstream libraries/users will know about these issues.

Related issue (same function): https://github.com/thestk/rtmidi/issues/373 

Everything after the break was reported by GitHub Copilot (Claude Opus 5)

Pete
Microsoft

---

## Summary

The System Exclusive path in `sendMessage` prepares a `MIDIHDR`, and on two of its three exits it releases the buffer without unpreparing that header first. The header is a stack local, so the function then returns and the stack frame is reused.

There is also an unbounded retry loop whose exit condition only recognizes one specific error, so every other error leaves the same problem.

## The code

[`RtMidi.cpp#L3161-L3190`](https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/RtMidi.cpp#L3161-L3190):

```cpp
MIDIHDR sysex{};
sysex.lpData = (LPSTR) buffer;
sysex.dwBufferLength = nBytes;
sysex.dwFlags = 0;
result = midiOutPrepareHeader( data->outHandle, &sysex, sizeof( MIDIHDR ) );
if ( result != MMSYSERR_NOERROR ) {
  free( buffer );                       // fine: prepare failed, nothing to undo
  ...
  return;
}

result = midiOutLongMsg( data->outHandle, &sysex, sizeof( MIDIHDR ) );
if ( result != MMSYSERR_NOERROR ) {
  free( buffer );                       // header is still prepared
  ...
  return;
}

while ( MIDIERR_STILLPLAYING == midiOutUnprepareHeader( data->outHandle, &sysex, sizeof ( MIDIHDR ) ) ) Sleep( 1 );
free( buffer );
```

Two things to credit before the defects: the `MIDIHDR` is value-initialized with `{}`, which is correct and is better than the input side of the same file, and the prepare-failure path is right.

## The two exits that are wrong

**`midiOutLongMsg` failure at [`L3178-L3183`](https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/RtMidi.cpp#L3178-L3183).** `midiOutPrepareHeader` succeeded, so the driver has been handed this header and the buffer behind it. The code frees the buffer and returns, so `sysex` goes out of scope still prepared and `buffer` is gone. Nothing ever calls `midiOutUnprepareHeader`. The correct sequence is to unprepare and only then free.

**Any non-`STILLPLAYING` error from `midiOutUnprepareHeader` at [`L3187`](https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/RtMidi.cpp#L3187).** The loop tests only for `MIDIERR_STILLPLAYING`. Any other non-zero result — an invalid handle after a device is removed mid-transfer, for instance — compares unequal, so the loop exits immediately, the result is discarded, and `free( buffer )` runs anyway. That is the same defect as the one above, reached by a different route, and it is the route I would expect a user to hit, because surprise device removal during a long dump is a real thing that happens to people sending firmware updates and patch banks.

**Suggestion:** capture the result, retry only on `MIDIERR_STILLPLAYING`, bound the retry, and treat every other value as a failure that is reported without freeing a buffer the driver may still hold.

## The retry loop has no bound

Same line. `while ( MIDIERR_STILLPLAYING == midiOutUnprepareHeader(...) ) Sleep( 1 );` has no iteration cap, no deadline and no cancellation.

I want to be careful here, because I have not made this loop spin and I am not claiming it hangs in practice. What I can say is that the exit condition is entirely in the hands of whichever WinMM driver is underneath, and a library does not get to choose that. Windows still supports third-party `.drv` MIDI drivers, and how eagerly any given one retires a buffer is not something RtMidi can probe for. Writing to the contract is correct everywhere; relying on the behavior of the drivers that happened to be on the development machine is correct until a user installs something else. For comparison, JUCE caps the equivalent retry rather than looping forever.

**Suggestion:** bound it, and surface a failure if the bound is reached.

## A related expectation, for the documentation rather than the code

`sendMessage` is synchronous for System Exclusive. `midiOutLongMsg` can block until the transfer has essentially completed, and over a DIN connection at 31250 baud a large dump takes as long as it takes. Callers sending firmware updates or patch banks from a user interface thread will see that as their application freezing. This is a property of the platform rather than an RtMidi defect, but it is worth a sentence in the API documentation, because the name `sendMessage` does not suggest it. There is a fuller description in [Microsoft's porting guidance for MIDI libraries](https://aka.ms/MidiLibraryPorting).

## `closePort` discards the `midiOutClose` result

[`L3101-L3113`](https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/RtMidi.cpp#L3101-L3113):

```cpp
// midiOutReset( data->outHandle );     // deliberately disabled, see #222
midiOutClose( data->outHandle );
data->outHandle = 0;
connected_ = false;
```

`midiOutClose` returns `MIDIERR_STILLPLAYING` if buffers are still queued, and in that case the handle is not closed. The result is discarded and the handle is overwritten with zero, so the handle is leaked and the object reports itself closed. Given that the `midiOutReset` which would have retired those buffers is deliberately commented out for the good reason in #222, this is more reachable here than it would otherwise be.

**Suggestion:** check the result. If it is `MIDIERR_STILLPLAYING`, either report it or wait and retry a bounded number of times before giving up, but do not drop the handle on the floor.

## The uninitialized `packet` DWORD is real but benign, and I measured it

[`L3193-L3199`](https://github.com/thestk/rtmidi/blob/a3233c22949342f6697681e2cf2403e27fcf0c9e/RtMidi.cpp#L3193-L3199):

```cpp
DWORD packet;
unsigned char *ptr = (unsigned char *) &packet;
for ( unsigned int i=0; i<nBytes; ++i ) { *ptr = message[i]; ++ptr; }
```

`packet` is uninitialized and only `nBytes` of its four bytes are written, so a one- or two-byte message passes the remaining bytes to `midiOutShortMsg` as whatever was on the stack.

I checked whether this actually corrupts anything, because I did not want to report it as more than it is. On Windows 11, through a USB MIDI interface with a physical DIN loopback, the unused bytes are discarded: a Program Change sent with `0xFFFF` in the unused bytes arrives byte-for-byte identical to the same message sent with zeros, and so does a single-byte Timing Clock sent with `0x7F7F7F` in the unused bytes. The documented contract agrees — `midiOutShortMsg` specifies the high byte as not used, and the middle bytes as present only if the message needs them.

So this does not put bad data on the wire. It is worth fixing only because passing an indeterminate value is undefined behavior that MSVC's `/analyze` and the sanitizers will flag, and `DWORD packet = 0;` costs nothing.

## Scope

Read from source at the commit above, on Windows 11. The only claim here I have measured is the last one, and the measurement is the reason I have de-escalated it rather than led with it. I have not reproduced the buffer-lifetime defects; they are read from the code.

The input side and the `__WINDOWS_UWP__` backend are reported separately.

insolace:
Five defects in the WinMM backend, all reachable when sending or receiving large SysEx — firmware updates and patch banks. Rebased onto current `master`.

## Output: SysEx routed by framing state, not length (#373)

`sendMessage()` chose between `midiOutLongMsg()` and `midiOutShortMsg()` on `nBytes > 3` alone. A caller pacing a large SysEx across several calls frequently ends with a short span — often a bare `F7` — and that tail took the short-message branch, which must not be used for SysEx. The remaining bytes never reached the device and the transfer never completed.

`WinMidiData` now tracks whether a SysEx is open, so a continuation is routed by what it is rather than by how big it happens to be. The new state is committed only once `midiOutLongMsg()` has accepted the bytes, so a failed send leaves it untouched and a retry of the same span routes identically.

## Output: buffer and header lifetime (#377)

- The `MIDIHDR` and its buffer are heap-allocated. Once `midiOutPrepareHeader()` succeeds the driver holds pointers to both, so any exit that may leave it owning them leaks both deliberately. Leaking only the buffer was not enough — a stack `MIDIHDR` is reused the moment the function returns.
- `midiOutLongMsg()` failure unprepares before freeing, and frees only if the unprepare succeeded.
- The unprepare retry is bounded by a 5 s deadline. Any non-`MIDIERR_STILLPLAYING` result is a failure rather than a reason to fall out of the loop and free a buffer the driver may still own — which is the path surprise device removal during a long dump takes.
- The bound is a `GetTickCount()` deadline rather than an iteration count: `Sleep(1)` sleeps for at least one scheduler tick, ~15.6 ms by default, so counting 1 ms iterations overshoots badly.
- `closePort()` no longer discards `midiOutClose()`'s result, and the short-message `DWORD` is value-initialized.

## Input: closePort teardown (#376, #325)

`MidiInWinMM::closePort()` could return part-way through with the critical section still held, `connected_` still true and the sysex buffers already freed, so the destructor ran the same loop again over deleted pointers.

Teardown now lives in one helper shared by `closePort()` and every `openPort()` failure path. Every step runs regardless of earlier failures, nothing returns early, a header the driver will not release is leaked rather than freed, the vector is cleared, and the first failure is reported after cleanup — as a warning, since `closePort()` also runs from the destructor where a thrown `RtMidiError` would terminate the process. The headers are `new MIDIHDR()` / `delete` rather than the mismatched `new char[]` / `delete[]`.

**Repro, without a firmware update:** hold an input port open on any device, disable it in Device Manager, then close the port. On `master` the process dies with `0xC0000409`; with this PR it reports `MIDIERR_STILLPLAYING` as a warning, leaks the buffer the driver still owns, and exits cleanly. Also note `master`'s message in that path says `MidiInWinMM::openPort:` while running in `closePort()`, which sends anyone reading a user's log to the wrong function; that is corrected here.

## Input: the reopen flag (side effect of #372)

#372 added `std::atomic<bool> closing`, set in `closePort()` and cleared only in `initialize()`. `openPort()` never clears it, so a second `openPort()` on the same `RtMidiIn` runs with it still set, `midiInputCallback()` never requeues, and SysEx input stops once the buffers queued by `openPort()` are consumed — short messages keep working. Measured on hardware: 8 identity requests per open gives 8/8 on the first open and 4/8 on every open after it, 4 being the default `bufferCount`. The shared teardown helper clears the flag, so this PR fixes it as a side effect.

## Open-path error codes (#234)

Five failure paths in `MidiInWinMM::openPort()` and `MidiOutWinMM::openPort()` had an `MMRESULT` in scope and discarded it, so callers got "error creating Windows MM MIDI input port" with nothing to act on. They now report the code, as @SpotlightKid asked for in that thread. This does not change what can be opened — the original report is Windows' exclusive-open behaviour (refs #102) — but it makes the difference between a busy device, a missing one and a driver fault visible.

## Credit

The deadlock between `closePort()` and `midiInputCallback()` was found first by @jkeller51 in #372, which is now merged. This PR previously carried its own fix for the same race; that has been dropped in favour of his, and the branch rebased on top.

## Verification

Builds clean on MSVC x64 and Win32 (`/W4`) and MinGW GCC 16.1.0 UCRT x86_64 (`-Wall -Wextra`), with upstream's test programs. Hardware checks on Windows 11 against a KMI 12 Step, WinMM backend, each run as an A/B against unmodified `master`:

- **#373:** a 6-byte identity request split 4+2, 3+3, 2+2+2 and 1×6 fails on `master` (`sendMessage` error, no reply) and passes on this branch. Whole-message and 5+1 splits pass on both — the bug needs a trailing span of 3 bytes or fewer, which is why it went unnoticed.
- **#376:** the Device Manager repro above — crash on `master`, clean exit here.
- **Reopen:** 4/8 sysex replies after the second `openPort()` on `master`, 8/8 here.

## Scope

Fixes #373. Fixes #376. Fixes #325. Fixes #377. Addresses #234. Refs #333 — the same unbounded unprepare loop, now bounded, but the follow-up there says a timeout alone did not stop the hang and part of that crash was likely #376, so this is not claimed as a fix.

The `midiInAddBuffer`-from-inside-the-callback redesign raised in #376 is left for separate work, as is #375 (UWP). Adds a documentation note that SysEx sends are synchronous on WinMM.

### DIFF
--- a/RtMidi.cpp
+++ b/RtMidi.cpp
@@ -2688,7 +2688,19 @@ struct WinMidiData {
   MidiInApi::MidiMessage message;
   std::vector<LPMIDIHDR> sysexBuffer;
   CRITICAL_SECTION _mutex; // [Patrice] see https://groups.google.com/forum/#!topic/mididev/6OUjHutMpEo
+  // Output-side SysEx framing state: true once 0xF0 has been sent and no 0xF7
+  // (or other terminating status byte) has been seen yet. Lets sendMessage()
+  // route a continuation span to midiOutLongMsg() regardless of its length.
+  bool sysexInProgress = false;
+  // Input-side teardown flag, guarded by _mutex: set while the sysex buffers
+  // are being retired so midiInputCallback() stops requeueing them.
+  bool inputClosing = false;
 };
+
+// How long to wait for a driver to release a buffer (or to close a port that
+// still has buffers queued) before giving up. Deliberately generous: this
+// guards against a driver that never lets go, it is not a tuning value.
+static const DWORD kWinMMBufferReleaseTimeoutMs = 5000;
 
 //*********************************************************************//
 //  API: Windows MM
@@ -2760,10 +2772,16 @@ static void CALLBACK midiInputCallback( HMIDIIN /*hmin*/,
     // buffer when an application closes and in this case, we should
     // avoid requeueing it, else the computer suddenly reboots after
     // one or two minutes.
-    if ( apiData->sysexBuffer[sysex->dwUser]->dwBytesRecorded > 0 ) {
-      //if ( sysex->dwBytesRecorded > 0 ) {
+    //
+    // Use the header WinMM handed back rather than indexing sysexBuffer: it is
+    // the same object, and the vector is cleared once the port has closed. Do
+    // not requeue while the port is closing, since the buffers are about to be
+    // unprepared and freed.
+    if ( sysex->dwBytesRecorded > 0 ) {
+      MMRESULT result = MMSYSERR_NOERROR;
       EnterCriticalSection( &(apiData->_mutex) );
-      MMRESULT result = midiInAddBuffer( apiData->inHandle, apiData->sysexBuffer[sysex->dwUser], sizeof(MIDIHDR) );
+      if ( !apiData->inputClosing )
+        result = midiInAddBuffer( apiData->inHandle, sysex, sizeof(MIDIHDR) );
       LeaveCriticalSection( &(apiData->_mutex) );
       if ( result != MMSYSERR_NOERROR )
         std::cerr << "\nRtMidiIn::midiInputCallback: error sending sysex to Midi device!!\n\n";
@@ -2788,6 +2806,52 @@ static void CALLBACK midiInputCallback( HMIDIIN /*hmin*/,
 
   // Clear the vector for the next input message.
   apiData->message.bytes.clear();
+}
+
+// Stop input and release everything MidiInWinMM::openPort() acquired: the
+// sysex buffers and the input handle. Used by closePort() and by openPort()'s
+// own failure paths, so it has to cope with a partly built buffer list.
+//
+// Every step runs regardless of earlier failures and nothing here throws or
+// returns early, so the port can never be left half torn down (see #376).
+// Returns the first failure, or MMSYSERR_NOERROR.
+static MMRESULT releaseWinMMInput( WinMidiData *data )
+{
+  // Stop the callback requeueing buffers before retiring them. The flag is
+  // set under the lock, but the lock is not held across midiInReset(), which
+  // drives the very callbacks that take it.
+  EnterCriticalSection( &(data->_mutex) );
+  data->inputClosing = true;
+  LeaveCriticalSection( &(data->_mutex) );
+
+  midiInReset( data->inHandle );
+  midiInStop( data->inHandle );
+
+  MMRESULT firstFailure = MMSYSERR_NOERROR;
+  for ( size_t i=0; i < data->sysexBuffer.size(); ++i ) {
+    if ( data->sysexBuffer[i] == NULL ) continue;
+    MMRESULT result = midiInUnprepareHeader( data->inHandle, data->sysexBuffer[i], sizeof(MIDIHDR) );
+    if ( result != MMSYSERR_NOERROR ) {
+      // The driver may still own this header, so leak it rather than free it.
+      if ( firstFailure == MMSYSERR_NOERROR ) firstFailure = result;
+      data->sysexBuffer[i] = NULL;
+    }
+  }
+
+  MMRESULT result = midiInClose( data->inHandle );
+  if ( result != MMSYSERR_NOERROR && firstFailure == MMSYSERR_NOERROR ) firstFailure = result;
+  data->inHandle = 0;
+
+  // No callbacks arrive once the handle is closed, so the headers can go now.
+  for ( size_t i=0; i < data->sysexBuffer.size(); ++i ) {
+    if ( data->sysexBuffer[i] == NULL ) continue;
+    delete [] data->sysexBuffer[i]->lpData;
+    delete data->sysexBuffer[i];
+  }
+  data->sysexBuffer.clear();
+  data->inputClosing = false;
+
+  return firstFailure;
 }
 
 MidiInWinMM :: MidiInWinMM( const std::string &clientName, unsigned int queueSizeLimit )
@@ -2868,7 +2932,7 @@ void MidiInWinMM :: openPort( unsigned int portNumber, const std::string &/*port
   // Allocate and init the sysex buffers.
   data->sysexBuffer.resize( inputData_.bufferCount );
   for ( unsigned int i=0; i < inputData_.bufferCount; ++i ) {
-    data->sysexBuffer[i] = (MIDIHDR*) new char[ sizeof(MIDIHDR) ];
+    data->sysexBuffer[i] = new MIDIHDR();
     data->sysexBuffer[i]->lpData = new char[ inputData_.bufferSize ];
     data->sysexBuffer[i]->dwBufferLength = inputData_.bufferSize;
     data->sysexBuffer[i]->dwUser = i; // We use the dwUser parameter as buffer indicator
@@ -2876,8 +2940,7 @@ void MidiInWinMM :: openPort( unsigned int portNumber, const std::string &/*port
 
     result = midiInPrepareHeader( data->inHandle, data->sysexBuffer[i], sizeof(MIDIHDR) );
     if ( result != MMSYSERR_NOERROR ) {
-      midiInClose( data->inHandle );
-      data->inHandle = 0;
+      releaseWinMMInput( data );
       errorString_ = "MidiInWinMM::openPort: error starting Windows MM MIDI input port (PrepareHeader).";
       error( RtMidiError::DRIVER_ERROR, errorString_ );
       return;
@@ -2886,8 +2949,7 @@ void MidiInWinMM :: openPort( unsigned int portNumber, const std::string &/*port
     // Register the buffer.
     result = midiInAddBuffer( data->inHandle, data->sysexBuffer[i], sizeof(MIDIHDR) );
     if ( result != MMSYSERR_NOERROR ) {
-      midiInClose( data->inHandle );
-      data->inHandle = 0;
+      releaseWinMMInput( data );
       errorString_ = "MidiInWinMM::openPort: error starting Windows MM MIDI input port (AddBuffer).";
       error( RtMidiError::DRIVER_ERROR, errorString_ );
       return;
@@ -2896,8 +2958,7 @@ void MidiInWinMM :: openPort( unsigned int portNumber, const std::string &/*port
 
   result = midiInStart( data->inHandle );
   if ( result != MMSYSERR_NOERROR ) {
-    midiInClose( data->inHandle );
-    data->inHandle = 0;
+    releaseWinMMInput( data );
     errorString_ = "MidiInWinMM::openPort: error starting Windows MM MIDI input port.";
     error( RtMidiError::DRIVER_ERROR, errorString_ );
     return;
@@ -2917,27 +2978,18 @@ void MidiInWinMM :: closePort( void )
 {
   if ( connected_ ) {
     WinMidiData *data = static_cast<WinMidiData *> (apiData_);
-    EnterCriticalSection( &(data->_mutex) );
-    midiInReset( data->inHandle );
-    midiInStop( data->inHandle );
-
-    for ( size_t i=0; i < data->sysexBuffer.size(); ++i ) {
-      int result = midiInUnprepareHeader(data->inHandle, data->sysexBuffer[i], sizeof(MIDIHDR));
-      delete [] data->sysexBuffer[i]->lpData;
-      delete [] data->sysexBuffer[i];
-      if ( result != MMSYSERR_NOERROR ) {
-        midiInClose( data->inHandle );
-        data->inHandle = 0;
-        errorString_ = "MidiInWinMM::openPort: error closing Windows MM MIDI input port (midiInUnprepareHeader).";
-        error( RtMidiError::DRIVER_ERROR, errorString_ );
-        return;
-      }
-    }
-
-    midiInClose( data->inHandle );
-    data->inHandle = 0;
+    MMRESULT result = releaseWinMMInput( data );
     connected_ = false;
-    LeaveCriticalSection( &(data->_mutex) );
+
+    // Reported only once the port is fully closed, and as a warning rather
+    // than an error: nothing is left for the caller to act on, and
+    // closePort() also runs from the destructor, where a thrown RtMidiError
+    // would terminate the process.
+    if ( result != MMSYSERR_NOERROR ) {
+      errorString_ = "MidiInWinMM::closePort: error closing Windows MM MIDI input port (MMRESULT " +
+                     std::to_string( result ) + ").";
+      error( RtMidiError::WARNING, errorString_ );
+    }
   }
 }
 
@@ -3106,9 +3158,29 @@ void MidiOutWinMM :: closePort( void )
     // Controllers" (to all 16 channels) CC messages which is undesirable (see issue #222)
     // midiOutReset( data->outHandle );
 
-    midiOutClose( data->outHandle );
+    // midiOutClose() returns MIDIERR_STILLPLAYING while buffers are still
+    // queued, and in that case the handle is NOT closed. The midiOutReset()
+    // that would retire them is deliberately disabled above, so give the
+    // driver a bounded time to finish rather than dropping the handle.
+    const DWORD start = GetTickCount();
+    MMRESULT result;
+    for ( ;; ) {
+      result = midiOutClose( data->outHandle );
+      if ( result != MIDIERR_STILLPLAYING || GetTickCount() - start >= kWinMMBufferReleaseTimeoutMs ) break;
+      Sleep( 1 );
+    }
     data->outHandle = 0;
+    data->sysexInProgress = false;
     connected_ = false;
+
+    // Reported once the object is back in a consistent state, and as a
+    // warning rather than an error: closePort() also runs from the
+    // destructor, where a thrown RtMidiError would terminate the process.
+    if ( result != MMSYSERR_NOERROR ) {
+      errorString_ = "MidiOutWinMM::closePort: error closing Windows MM MIDI output port (MMRESULT " +
+                     std::to_string( result ) + "); the handle may have leaked.";
+      error( RtMidiError::WARNING, errorString_ );
+    }
   }
 }
 
@@ -3148,11 +3220,46 @@ void MidiOutWinMM :: sendMessage( const unsigned char *message, size_t size )
 
   MMRESULT result;
   WinMidiData *data = static_cast<WinMidiData *> (apiData_);
-  if ( nBytes > 3 ) { // Sysex message
 
-    // Allocate buffer for sysex data.
+  // Decide whether this buffer belongs on the long-message (SysEx) path.
+  //
+  // Routing on length alone is wrong: a caller may hand a large SysEx to
+  // sendMessage() in several spans so it can pace the transfer, and the final
+  // span is frequently short - often a bare 0xF7. Under a size-only test that
+  // tail took the midiOutShortMsg() branch, which must not be used for SysEx,
+  // so the remaining bytes never reached the device and the transfer never
+  // completed. Track the SysEx state instead, so a continuation is routed by
+  // what it is rather than by how big it happens to be.
+  bool isSysex = data->sysexInProgress || message[0] == 0xF0;
+  if ( !isSysex && nBytes > 3 ) isSysex = true;
+
+  if ( isSysex ) {
+
+    // Work out the framing state this buffer leaves behind, but commit it only
+    // once midiOutLongMsg() has accepted the bytes (below). If the send fails,
+    // the state stays as it was, so a caller that retries the same span gets
+    // it routed the same way; otherwise a retried tail of 1-3 bytes would drop
+    // to midiOutShortMsg() and be lost again. Real-time bytes (0xF8-0xFF) may
+    // be interleaved mid-SysEx without ending it; 0xF7 ends it, and so does
+    // any other non-real-time status byte.
+    bool nextSysexInProgress = data->sysexInProgress;
+    for ( unsigned int i=0; i<nBytes; ++i ) {
+      const unsigned char b = message[i];
+      if ( b >= 0xF8 ) continue;
+      if ( b == 0xF0 ) { nextSysexInProgress = true; continue; }
+      if ( b == 0xF7 ) { nextSysexInProgress = false; continue; }
+      if ( b >= 0x80 ) nextSysexInProgress = false;
+    }
+
+    // Allocate the MIDIHDR and its buffer on the heap. Once
+    // midiOutPrepareHeader() succeeds the driver holds pointers to both, so
+    // the failure exits below that may leave the driver owning them leak both
+    // on purpose. Leaking only the buffer is not enough: a stack MIDIHDR would
+    // be reused as soon as this function returned (#377).
+    MIDIHDR *sysex = new MIDIHDR();
     char *buffer = (char *) malloc( nBytes );
     if ( buffer == NULL ) {
+      delete sysex;
       errorString_ = "MidiOutWinMM::sendMessage: error allocating sysex message memory!";
       error( RtMidiError::MEMORY_ERROR, errorString_ );
       return;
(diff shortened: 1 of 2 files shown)


insolace:
Additional hardware evidence for the #373 half of this PR, from a same-machine A/B run while testing an unrelated branch.

Two builds on one Windows 11 machine, minutes apart, against a KMI SoftStep 3 and a QuNexus. The first is a branch based on current `master`, so it has upstream's WinMM and none of the fixes here; the second is the same tree with this PR's changes applied. A universal identity request was sent whole, and split across `sendMessage()` calls in five patterns:

| split pattern | without this PR | with this PR |
|---|---|---|
| whole message | pass | pass |
| 5 + 1 (trailing bare `F7`) | pass | pass |
| 4 + 2 | **fail** | pass |
| 3 + 3 | **fail** | pass |
| 2 + 2 + 2 | **fail** | pass |
| 1 × 6 | **fail** | pass |

Same result on both devices. Without the fix the failing cases report `MidiOutWinMM::sendMessage: error sending MIDI message` and the identity reply never arrives; with it, all six patterns complete.

That is #373 reproducing on real hardware and this PR fixing it, independent of the fault-injection suite described in the PR body. Worth noting the failure needs a trailing span of 3 bytes or fewer to be interesting — a 5+1 split happens to pass here, so a caller whose chunking never lands on a short tail would not see the bug at all, which is probably why it went unnoticed for so long.

Unrelated but visible in the same testing: the Windows MIDI Services backend is now proposed separately in #382. That branch deliberately excludes these WinMM changes, which is why it still shows the failures above — the two are independent and can land in either order.

insolace:
Credit where it is due: #372 found this deadlock first, back in May, and fixes it the same way — a flag set before teardown that stops `midiInputCallback` requeueing buffers. We hit it independently while working on #376 and did not notice the overlap until now.

Two differences. #372 is 18 lines and does only this, which makes it much easier to review than what is here. Ours sets the flag under the lock rather than using an atomic, and uses the header WinMM hands back instead of indexing `sysexBuffer`, since that vector gets cleared during close.

If you would rather take #372 for the deadlock and have us rebase this on top, that is fine by us.

Psychlist1972:
> on closePort(): when midiOutClose() returns MIDIERR_STILLPLAYING we sleep and retry for up to 5 s, on the basis that a failed close leaves the handle valid. Is that safe across legacy drivers and the Windows MIDI Services WinMM compatibility layer, or can a close report STILLPLAYING and still invalidate the handle? If retrying is not appropriate, would you recommend reporting immediately instead?

Thanks for the note on reproducing 376 with hardware.

> Sleep(1) sleeps for at least one scheduler tick, ~15.6 ms by default, so counting 1 ms iterations overshoots badly. Measured at 4.99 s over 312 retries in a fault-injection build.

Good article on that behavior here: https://randomascii.wordpress.com/2020/10/04/windows-timer-resolution-the-great-rule-change/

It's also why I added two functions to `MidiClock` early on: `BeginLowLatencySystemTimerPeriod` and the end function. User mode code can only request down to 1ms, but kernel driver code can request half a millisecond. Most folks running low-latency audio driver will already have half a millisecond system-wide. But you can't count on that, and there's some complexity to how it works, as explained in that blog post.

The rest via GitHub Copilot (Claude Opus 5) :

Retrying is correct, and the documentation is explicit about why.

On the `hmo` parameter, [`midiOutClose`](https://learn.microsoft.com/windows/win32/api/mmeapi/nf-mmeapi-midioutclose) says *"If the function is successful, the handle is no longer valid after the call to this function."* Invalidation is conditioned on success. The Remarks then say that pending `midiOutLongMsg` buffers mean *"the close operation will fail"*. A close that reports `MIDIERR_STILLPLAYING` did not close, and the handle is still yours. That holds for the WinMM compatibility layer too — it is implementing the same contract.

Two things worth adding rather than changing:

The documented remedy for pending buffers is `midiOutReset`, and you deliberately cannot use it — it is commented out in `closePort` because of #222, since it emits All Notes Off and Reset All Controllers across all 16 channels. Your bounded retry is the right substitute given that constraint, but it is worth a comment at the retry site saying so, or someone will eventually "fix" it by restoring the `midiOutReset` call and reintroduce #222.

You are also safe even if the answer had gone the other way. If some driver did invalidate the handle on a failed close, the next call returns `MMSYSERR_INVALHANDLE`, which is a documented return value — WinMM validates the handle rather than dereferencing it. The loop degrades to a harmless error rather than a use-after-free, so you did not need certainty here to be safe.

I would not switch to reporting immediately. The case you are protecting against is a large dump still draining, which is exactly the firmware-update path this PR is fixing, so an immediate report would surface a spurious failure precisely where it hurts.

---

Pete
Microsoft

insolace:
Rebased and reworked. The PR body above has been rewritten to match; the summary of what changed:

**Rebased onto current `master` (`2bed2e0`)**, so this is no longer conflicting. #372, #374, #378, #389 and #390 have all landed since the branch was cut.

**Our duplicate deadlock fix is gone.** @jkeller51 found that race first in #372, which is now merged, and this branch is rebased on top of it rather than carrying a second fix for the same thing. The teardown helper here clears his `closing` flag at the end of teardown, which turns out to matter — see below.

**@Psychlist1972's answer is now recorded in the source**, at the retry site, rather than living only in this thread: that `midiOutReset` stays disabled deliberately because of #222 and must not be "fixed" back in; that the bound is a `GetTickCount()` deadline rather than an iteration count because `Sleep(1)` sleeps for at least a scheduler tick; and both halves of the safety argument — `midiOutClose()` only invalidates the handle on success, and even if a driver got that wrong, WinMM validates handles so the loop would degrade to `MMSYSERR_INVALHANDLE` rather than a use-after-free. Thank you for the detail, it is exactly the kind of thing that gets lost.

**New in this revision: a defect #372 left behind.** `closing` is set in `closePort()` and cleared only in `initialize()`, so a second `openPort()` on the same `RtMidiIn` runs with it still set, `midiInputCallback()` never requeues, and SysEx input stops once the buffers queued by `openPort()` are consumed — while short messages keep working. On hardware, 8 identity requests per open: 8/8 on the first open, 4/8 on every open after it, 4 being the default `bufferCount`. Fixed here as a side effect of the shared teardown helper.

**The open-path error codes (#234) are absorbed into this PR.** Marked as "Addresses" rather than "Fixes": that reporter's port could not be opened because another application held it (refs #102), which this does not change, but it delivers what the thread actually asked for — the `MMRESULT` in the message.

**Verification is narrower and all first-hand.** The previous body cited QuNexus and SoftStep results and a fault-injection suite from an earlier machine; those were for the pre-rebase commit and are gone. What is claimed now was run against this tip: MSVC x64 and Win32 `/W4`, MinGW GCC 16.1.0 UCRT `-Wall -Wextra`, and a KMI 12 Step over WinMM, each hardware check as an A/B against unmodified `master`. All five CI legs are green, including both MinGW targets.

The #376 repro is now much easier for anyone to run: hold an input port open, disable the device in Device Manager, close the port. `master` dies with `0xC0000409`; this branch reports `MIDIERR_STILLPLAYING` as a warning and exits cleanly. Worth noting `master`'s message in that path names `MidiInWinMM::openPort` while executing in `closePort()`, which is corrected here.

@Psychlist1972 — anything further from your side before this is merged?

Psychlist1972:
The previous version of that comment had some more questions for me. I assume those are sorted now.

I haven't tested anything here, but if you covered the main issues, this seems fine. We're still working on what the long-term reconnect-within-the service is going to look like, but we're not close to a good universal solution on that yet.

Other than that caveat, it all seems good.

Pete
Microsoft

Issue #307, opened 2023-01-21
Does rtmidi support midi 2.0?


jwlrs:
Since I don't see anything explicit about midi 2.0 I am assuming it is unsupported, but wanted to ask.

insolace:
[...]

PR #382, opened 2026-09-12
Add a Windows MIDI Services backend

insolace:
# Add a Windows MIDI Services backend

Adds `RtMidi::WINDOWS_MIDI_SERVICES`, a backend that sends and receives MIDI 1.0 through [Windows MIDI Services](https://github.com/microsoft/MIDI) (the `Microsoft.Windows.Devices.Midi2` SDK). **It is opt-in:** nothing changes unless you configure with `-DRTMIDI_API_WMS=ON`, and it needs the Windows MIDI Services SDK runtime installed.

References #307 but does not close it. This bridges MIDI 1.0 over UMP; it is not MIDI 2.0 support.

## Prerequisites, and why both backends belong in one build

Worth stating plainly, because it drives several decisions below and it caught us
out when we first shipped against this SDK. @Psychlist1972 — corrections welcome
if any of this is wrong or has since changed; this is our experience rather than
anything authoritative.

There are three separate things, and they are easy to conflate:

- **Windows MIDI Services itself** — the service. Present on retail Windows 11
  24H2 and later. Nothing to install.
- **The Desktop App SDK** — what a *developer* needs to build against this
  backend: the `.winmd` for C++/WinRT projection generation, MSVC, C++17. Source
  and releases at <https://github.com/microsoft/MIDI>.
- **The Desktop App SDK runtime** — what an *end user* needs in order to run an
  application built against it. This is a separate download today, from
  <https://microsoft.github.io/MIDI/>. It is not the same thing as the service
  shipping in Windows.

That last distinction is the one that bit us. On a machine with the service but
without the SDK runtime, the backend's bootstrapper fails at
`CoCreateInstance()` with `0x80040154` (class not registered) before any port is
enumerated. The download page presents the runtime alongside the MIDI Settings
app, console and PowerShell tooling, which reads as optional power-user extras;
for a third-party application it is not optional.

**The consequence for anyone shipping an application:** you cannot ship a
WMS-only binary and expect it to start on a user's machine. What works is one
build with WMS *and* WinMM compiled in, deciding at startup which to use, and
pointing the user at the installer if they want the newer stack. Three things in
this PR exist for that:

- WMS is listed before WinMM so `UNSPECIFIED` prefers it and falls back
  automatically when the runtime is absent (§2).
- `checkApiAvailability()` answers "can this build use this API on this
  machine?", and returns the installer URL when the answer is a missing runtime
  — which is how an application tells a user what to install. Note that
  availability and "has ports" are separate questions: WMS can be available with
  no endpoints attached, so an application should choose its API first and
  enumerate second rather than reading an empty port list as unavailability.
- A missing runtime degrades rather than throws: the port object lists no ports
  and `openPort()` reports `DRIVER_NOT_INSTALLED`, so an application can detect
  the situation and guide the user instead of failing at construction.

One related observation from our own testing, offered in case it is useful rather
than as a claim about the platform: on the Windows 11 machines we tested, WinMM
traffic appears to reach devices through a Windows MIDI Services compatibility
layer rather than a native driver, and it does not behave identically to one. A
transient device-busy condition that a native driver reported as
`MIDIERR_NOTREADY` came back as plain `MMSYSERR_ERROR` through that path. It did
not change what this PR does, but it is another reason an application benefits
from knowing which stack it is actually on.

## What it adds

- **Ports:** `MidiInWinMidi2` / `MidiOutWinMidi2` expose each endpoint's group terminal blocks as RtMidi ports, one per group when a block spans several.
- **Sending:**
  - MIDI 1.0 bytes are encoded to UMP by a stateful encoder, so running status and SysEx split across `sendMessage()` calls encode correctly.
  - A send is split at whole packets when it exceeds the service's per-send word limit, and the encoder's state advances only once every word has been sent, so a caller may retry a message that failed.
- **Receiving:**
  - SysEx7 fragments are reassembled into one message.
  - Only packets with a MIDI 1.0 form, on the port's own group, are delivered. Utility messages and MIDI 2.0 channel voice messages are not translated.
  - Timestamps are deltas derived from `MidiClock::TimestampFrequency()`.
- **Enumeration:** port lists come from a `MidiEndpointDeviceWatcher` cache that follows endpoints as they are added, updated and removed, and are refreshed when a port is opened.
- **Send size:** the number of UMP words per send is queried from the service (`GetSupportedMaxMidiWordsPerTransmission()`), falling back to 684 words on runtimes that predate the query.
- **Runtime detection:** `RtMidi::checkApiAvailability()` reports whether an API can be used by this build on this system, with an installer URL for WMS. Without the runtime a port object simply lists no ports, and `openPort()` reports `RtMidiError::DRIVER_NOT_INSTALLED`.
- **C API:** `RTMIDI_API_WINDOWS_MIDI_SERVICES` and `RTMIDI_ERROR_DRIVER_NOT_INSTALLED`.
- **Build:**
  - `RTMIDI_API_WMS` (default `OFF`), plus `cmake/FindCppWinRT.cmake` and `cmake/FindWindowsMidiServices.cmake`.
  - The SDK's C++/WinRT projection is generated from its `.winmd` at build time.
  - Enabling WMS raises the C++ standard to 17 and requires the MSVC toolset.
  - An optional `RTMIDI_USE_WMS_COM_RAW` selects the COM raw send path.

## Design notes, and things worth your opinion

1. **COM apartments — the backend never initializes COM on a thread the application owns.** It keeps the process's multi-threaded apartment alive with `CoIncrementMTAUsage()`, so threads that have not initialized COM use it implicitly, while an application thread can still be an STA — a GUI thread that later calls `OleInitialize()`, for example.

   This is worth stating early because @Psychlist1972 reported the opposite pattern in #375: an earlier version of this backend initialized an apartment from a static initializer. That is deliberately gone, and a regression test covers it. The suggestion on #375 was to initialize lazily on a thread RtMidi owns; `CoIncrementMTAUsage()` means no owned thread is needed at all. Measured here: with the old code the main thread was already MTA before `main()` and a later `OleInitialize()` returned `RPC_E_CHANGED_MODE`, so a GUI thread lost OLE. With this code the main thread is untouched and `OleInitialize()` returns `S_OK`.
2. **Default backend order changes when WMS is compiled in.** WMS is listed before WinMM, so `UNSPECIFIED` picks WMS when its runtime is present and falls back to WinMM when it is not. Nothing changes for builds without `RTMIDI_API_WMS=ON`. If you would rather WMS came after WinMM, or were only used when asked for explicitly, that is a one-line change.
3. **The endpoint watcher is started by the first port object and then runs for the life of the process,** so later objects do not wait for enumeration (about half a second here). During static destruction its handlers are revoked and the watcher itself is deliberately leaked. Releasing it there runs the system `DeviceWatcher`'s destructor, whose `CoDecrementMTAUsage()` fails fast under the loader lock: unloading a DLL that links RtMidi crashed the host until it was changed.
4. **`closePort()` from the input callback is supported.** Disconnecting waits for the SDK's receive thread, which is the thread running the callback, so the teardown is handed to another thread and finishes once the callback returns. A close from any other thread waits for a callback in progress before dropping the connection. Closing from the callback and from another thread at the same time is not supported.
5. **`sendMessage()` returns once the service accepts the message, not once it is on the wire,** unlike `MidiOutWinMM::sendMessage()`. A caller pacing a large chunked transfer gets no backpressure from this backend. When its buffers are full the service rejects a send rather than blocking.
6. **CMake only.** `configure.ac` has no WMS support, and neither do the bundled Visual Studio projects in `msw/`. Autotools on Windows means MinGW, which cannot compile the C++/WinRT this backend needs, and the build has to run `cppwinrt.exe` first — so Autotools support would be unusable in practice. Happy to add either if you want them.
7. **Not included:** a `vcpkg.json` manifest. An earlier version of this backend used vcpkg packages; the CMake modules above find the Windows SDK's `cppwinrt` and the installed WMS runtime directly, so the manifest is unused.
8. **Left alone:** `rtmidi_compiled_apis[]` lists `WEB_MIDI_API` twice. That is already in `master` and is unrelated to this change.

## Testing

The Windows work below is on Windows 11 25H2, MSVC 19.44, Windows MIDI Services
SDK runtime 1.0.17-rc.4. The Linux and macOS results are from two other machines, 
checking that a backend they cannot compile stays cleanly out of their builds.

**Builds — Windows**
- `RTMIDI_API_WMS=OFF` (the default): MSVC x64 at `/W4` builds clean with the test programs; `apinames` passes.
- `RTMIDI_API_WMS=ON`: MSVC x64 builds clean as a DLL and as a static library; `apinames` passes with `winmidi2` in both the C++ and C tables; `midiprobe` lists the WMS endpoints.
- MinGW GCC 11.2 x86_64 and GCC 8.1 i686 with WMS absent: `RtMidi.cpp` and `rtmidi_c.cpp` build clean with `-Wall -Wextra`; `apinames` and `midiprobe` link and run.

**Builds — WMS absent, independently verified on two other platforms**
- **Linux** (ALSA + JACK), CMake and Autotools: configure, build and `make check` pass in both. `apinames` exits 0 with WMS absent from the C++ *and* C tables, `testcapi` passes, and `midiprobe` lists only ALSA and JACK — no blank name and no phantom WMS row.
- **macOS** (26.4.1 arm64, AppleClang 21), CMake: configures as `coremidi` only and builds clean. `apinames`, `testcapi` and `midiprobe` show no WMS row, and grepping their combined output for `wms` / `windows_midi_services` finds nothing. CoreMIDI itself enumerates and probes live ports in the same run, so the absence is meaningful rather than a broken build. Autotools was not exercised there (no autoconf/automake on that machine).

  Absence is also structural rather than only observed: `RTMIDI_API_WMS` defaults `OFF` with no `WIN32` gate, forcing it on off-Windows is a `FATAL_ERROR`, and every WMS path is inside `#if defined(__WINDOWS_MIDI_SERVICES__)`, which that option alone defines.

**Windows MIDI Services loopback endpoints**
- SysEx payloads of 0, 1, 5, 6, 7, 12, 13, 100, 4104, 4105, 10000 and 65536 bytes, byte-exact round trip.
- One `sendMessage()` mixing channel messages, a 6000-byte SysEx, a real-time byte and a second SysEx, crossing the service's per-send word limit.
- A real-time byte inside a SysEx, SysEx split across many `sendMessage()` calls, and running status.
- Raw UMP injected into the loopback: utility, MIDI 2.0 channel voice, and other-group packets are not delivered; MIDI 1.0 channel voice, system real-time and SysEx7 are.
- 200 open/close/delete cycles under continuous traffic with AddressSanitizer, including `closePort()` from inside the callback, with no reports.
- Endpoints added and removed while the library is running, seen immediately in the port lists.
- Apartments: a thread with no COM, an STA thread, an MTA thread, and `OleInitialize()` after RtMidi use on the same thread.
- Timestamps against the sender's clock, process exit with ports open, a DLL that links RtMidi being unloaded and then endpoints changing, and a simulated missing runtime (falls back to WinMM, `openPort()` reports `DRIVER_NOT_INSTALLED`).
- The same SysEx and UMP tests with `RTMIDI_USE_WMS_COM_RAW=ON`.

**Hardware** (KMI SoftStep 3 and QuNexus)
- Enumerate, open, and send/receive over WMS.
- Every universal identity request got its identity reply, 3 of 3 attempts for each case: sent whole, and split into spans of 5+1 (a bare `F7`), 4+2, 3+3, 2+2+2 and 1×6. The split cases exercise SysEx encoding across calls.
- Firmware updates (chunked >60kb payloads) work on both devices with WinMM and WMS backends


insolace:
A worked example of the pattern described under "Prerequisites", in case it is useful — choosing between Windows MIDI Services and WinMM at startup, in one build that contains both.

The point it demonstrates: `getCompiledApi()` says what the *build* contains, `checkApiAvailability()` says what *this machine* can actually use, and those are different questions. When the answer is a missing runtime, `installUrl` is what an application shows the user.

Deliberately not added to the PR as a test program — it is Windows-and-MSVC-only, could not run in CI, and the PR is large enough. Posting it here instead.

```cpp
// Choosing between Windows MIDI Services and WinMM at startup.
//
// Build this with both backends compiled in:
//   cmake -S . -B build -DRTMIDI_API_WMS=ON -DRTMIDI_API_WINMM=ON
//
// The point: the Windows MIDI Services *service* ships in Windows 11 24H2+,
// but the Desktop App SDK *runtime* an application links against is a separate
// download. A machine can have the service and not the runtime, so a shipping
// application needs both backends compiled in, has to pick at startup, and
// should tell the user where to get the runtime if they want the newer stack.
//
// checkApiAvailability() answers that question without opening any ports.

#include "RtMidi.h"
#include <iostream>
#include <vector>

// Pick the best available API, preferring WMS, falling back to WinMM.
// Returns UNSPECIFIED if neither is usable.
static RtMidi::Api chooseApi()
{
  const RtMidi::Api preferred[] = { RtMidi::WINDOWS_MIDI_SERVICES,
                                    RtMidi::WINDOWS_MM };

  for ( RtMidi::Api api : preferred ) {
    RtMidi::RtMidiApiAvailability a = RtMidi::checkApiAvailability( api );

    std::cout << RtMidi::getApiDisplayName( api ) << ": "
              << ( a.available ? "available" : "unavailable" );
    if ( !a.available && !a.message.empty() )
      std::cout << " - " << a.message;
    std::cout << "\n";

    // A missing runtime is the actionable case: tell the user where to get it.
    if ( !a.available && !a.installUrl.empty() )
      std::cout << "  install it from: " << a.installUrl << "\n";

    if ( a.available )
      return api;
  }

  return RtMidi::UNSPECIFIED;
}

int main()
{
  // What this build actually contains, before asking what this machine has.
  std::vector<RtMidi::Api> compiled;
  RtMidi::getCompiledApi( compiled );
  std::cout << "Compiled APIs:";
  for ( RtMidi::Api api : compiled )
    std::cout << " " << RtMidi::getApiName( api );
  std::cout << "\n\n";

  const RtMidi::Api api = chooseApi();
  if ( api == RtMidi::UNSPECIFIED ) {
    std::cerr << "\nNo usable MIDI API on this machine.\n";
    return 1;
  }
  std::cout << "\nUsing " << RtMidi::getApiDisplayName( api ) << "\n";

  // From here on it is ordinary RtMidi: the chosen API is passed to the
  // constructor, and nothing else in the application needs to care which
  // backend it got.
  try {
    RtMidiOut out( api, "wmsfallback" );

    const unsigned n = out.getPortCount();
    std::cout << n << " output port(s):\n";
    for ( unsigned i = 0; i < n; ++i )
      std::cout << "  " << i << ": " << out.getPortName( i ) << "\n";

    if ( n == 0 ) {
      std::cout << "Nothing to open.\n";
      return 0;
    }

    // An individual port can refuse to open - it may be in use, or its driver
    // may simply reject it - so fall through to the next one.
    unsigned port = n;
    for ( unsigned i = 0; i < n && port == n; ++i ) {
      try {
        out.openPort( i );
        port = i;
      }
      catch ( RtMidiError & ) {
      }
    }
    if ( port == n ) {
      std::cerr << "No output port could be opened.\n";
      return 1;
    }

    // Universal Identity Request - harmless to any device that ignores it.
    std::vector<unsigned char> idRequest = { 0xF0, 0x7E, 0x7F, 0x06, 0x01, 0xF7 };
    out.sendMessage( &idRequest );
    std::cout << "Sent an identity request to port " << port
              << ": " << out.getPortName( port ) << "\n";
  }
  catch ( RtMidiError &error ) {
    // DRIVER_NOT_INSTALLED is reachable if the runtime disappears between the
    // availability check and the open, or if an application skips the check.
    error.printMessage();
    return 1;
  }

  return 0;
}
```

Built with `-DRTMIDI_API_WMS=ON -DRTMIDI_API_WINMM=ON`. Output with the SDK runtime present:

```
Compiled APIs: winmidi2 winmm

Windows MIDI Services: available

Using Windows MIDI Services
...
```

And with the runtime absent:

```
Compiled APIs: winmidi2 winmm

Windows MIDI Services: unavailable - The Windows MIDI Services SDK runtime is not
installed, or the MIDI service is not running.
  install it from: https://aka.ms/MidiServicesLatestSdkRuntimeInstaller
Windows MultiMedia: available

Using Windows MultiMedia
```

Two notes on the code. It opens the first port that will accept an open rather than assuming port 0, because a machine's first output port is often a synth that refuses one — on the test machine that is the GS Wavetable Synth, which fails a plain Win32 `midiOutOpen` with `MMSYSERR_ERROR` before RtMidi is involved. And it enumerates only after choosing an API, since an API can be available with no endpoints attached.

Unlike the rest of this PR, this example has been built and run on one machine only.

Psychlist1972:
@insolace this is using the older out-of-band SDK which will no longer be available come November. In November, Windows.Devices.Midi2 is in-box in Windows 11 25h2 and later.

24h2 will not get the new API, and we recommend customers on 24h2, which goes out of support next month, either move to 25h2 or turn on legacy mode.

An easy way to check for the API's presence is to try to `RoActivateInstance(L"Windows.Devices.Midi2.MidiApi")`. You'll get `E_NOINTERFACE` if that API is not present.

Then, if you have that returning `S_OK`, you need to check to see what `MidiApi.EnsureServiceAvailable()` returns. The reason is, on a system with `Windows.Devices.Midi2` present, if the user has selected to use Legacy API Mode, then the activate will succeed but the service will never start up.

Relevant docs:
- `Windows.Devices.Midi2.MidiApi`: https://microsoft.github.io/MIDI/sdk-reference/MidiApi/
- API Mode: https://microsoft.github.io/MIDI/kb/how-to-change-api-mode/

I recommend you point your agent to the samples here: https://github.com/microsoft/MIDI/tree/main/samples/cpp-winrt and to the docs above.

Pete
Microsoft

Psychlist1972:
Here's a sample showing how to detect the new API and service.

https://github.com/microsoft/MIDI/blob/main/samples/cpp-winrt/detect-midi-services/main_detect_midi_services.cpp

Pete
Microsoft

insolace:
Thanks @Psychlist1972, two questions:

1. What happens to applications already shipped against the out-of-band runtime after November? We have editors in customers' hands today that were built against Microsoft.Windows.Devices.Midi2 and rely on the end user having installed the SDK runtime from the download page. Does no longer available mean no longer distributed, with existing installs continuing to work, or does the runtime stop functioning on updated Windows? Those have very different consequences for us and we would rather plan than guess.

2. Is the in-box namespace API-compatible with the out-of-band one beyond the rename? Every type we use exists by name in both: MidiSession, MidiEndpointConnection, MidiEndpointDeviceWatcher, MidiEndpointDeviceInformation, MidiGroupTerminalBlock, MidiMessageReceivedEventArgs, MidiClock, and the calls Open, Close, CreateSession, CreateEndpointConnection, DisconnectEndpointConnection, SendMultipleMessagesWordArray, FillWords, GetSupportedMaxMidiWordsPerTransmission. Are the shapes the same, or is there porting work beyond changing the namespace and dropping the bootstrapper?

Psychlist1972:
> Thanks @Psychlist1972, two questions:
> 
> 1. What happens to applications already shipped against the out-of-band runtime after November? We have editors in customers' hands today that were built against Microsoft.Windows.Devices.Midi2 and rely on the end user having installed the SDK runtime from the download page. Does no longer available mean no longer distributed, with existing installs continuing to work, or does the runtime stop functioning on updated Windows? Those have very different consequences for us and we would rather plan than guess.

I know of a couple which have, and I have contacted those developers. The terms on the SDK did *not* allow shipping a production app. Similarly, the current previews do not have any go-live licenses for most partners, but they are also less impactful because they are not a central install for the previews. They are app-local.

> 2. Is the in-box namespace API-compatible with the out-of-band one beyond the rename? Every type we use exists by name in both: MidiSession, MidiEndpointConnection, MidiEndpointDeviceWatcher, MidiEndpointDeviceInformation, MidiGroupTerminalBlock, MidiMessageReceivedEventArgs, MidiClock, and the calls Open, Close, CreateSession, CreateEndpointConnection, DisconnectEndpointConnection, SendMultipleMessagesWordArray, FillWords, GetSupportedMaxMidiWordsPerTransmission. Are the shapes the same, or is there porting work beyond changing the namespace and dropping the bootstrapper?

There is some porting work beyond just namespace names, but it's generally minor if you are using the core Session/Connection API. There are a number of types which were changed from being structs to being runtimeclasses, so that means the accessors and setters are function calls not fields. "Enumeration" was added as a sub-namespace, and "Legacy" below that for all the MIDI 1.0 port-based enumeration help. The Watchers also now return the full type, not just an id, on add/update/remove, which was a common request from partners.

There's more if you use loopbacks or other transport configuration. Ask Copilot to summarize for you the full delta from each of the In-box dev previews to date. The first release was in July, announced in advance on the Discord server, but I know you don't use Discord much. All the important info is in the release notes:

1 https://github.com/microsoft/MIDI/releases?page=2#release-inbox-dev-preview-1
2 https://github.com/microsoft/MIDI/releases#release-inbox-dev-preview-2
3 https://github.com/microsoft/MIDI/releases#release-inbox-dev-preview-3
4 https://github.com/microsoft/MIDI/releases#release-inbox-dev-preview-4
5 https://github.com/microsoft/MIDI/releases#release-inbox-dev-preview-5
6 https://github.com/microsoft/MIDI/releases#release-inbox-dev-preview-6
7 https://github.com/microsoft/MIDI/releases#release-inbox-dev-preview-7
8 is coming this weekend

The SDK documentation I linked to is the current info for `Windows.Devices.Midi2`. Samples in the repo are also up to date with that.

Pete
Microsoft

Psychlist1972:
There's the comment in there about name prefixes. 

Port names are hard. Matching on port names, especially now that customers can change them to anything they want, is going to present real challenges. The new API has much better information which can be used to identify devices, endpoints, and their blocks/ports in a more robust way

This is what will ship in November and how we derive it. I hope this will help. 
https://microsoft.github.io/MIDI/kb/how-midi1-port-names-are-generated/

This is an app-centric view of moving concepts from WinMM to WMS
https://microsoft.github.io/MIDI/kb/moving-from-winmm-to-wms/

Here's a library-centric view:
https://microsoft.github.io/MIDI/kb/porting-midi-libraries/

Pete
Microsoft

insolace:
Putting a note here that I need to add virtual ports when this PR is updated

